PR #10, opened 2024-07-11
Migrate `ci-configuration` -> `fxci-config`

ahal:


ahal:
Despite the failures, this is ready to land:

* deploy / check actions are failing because https://phabricator.services.mozilla.com/D216298 hasn't landed yet. But that phab revision is failing because this hasn't landed yet. A nice little catch 22. So I'll land this first, then I'll land the phab revision afterwards.

* I'm not 100% sure why the openssf scorecard action is failing, but I suspect it's because it requires the workflow to be defined on the main branch, so hoping landing this will resolve that too. If not, I think it's fine to sort that out after the fact.

PR #11, opened 2024-07-11
Bug 1887949 - Change 'fxci-config' default branch back to 'main', r?#…

ahal:
…releng-reviewers!

Differential Revision: https://phabricator.services.mozilla.com/D216298


ahal:
This was already r+'ed in https://phabricator.services.mozilla.com/D216298, so going to bypass review here.

PR #13, opened 2024-07-11
ci: fix the deploy / check job on PRs

ahal:
It turns out PRs can't access the env defined in the Github environment, so we need to dynamically write the proper value to the $GITHUB_ENV file instead.


JohanLorenzo (approved):
LGTM!

PR #15, opened 2024-07-11
bug 1907211: add a single translations GPU worker in europe as a trial

bhearsum:
Let's see how these perform, and the actual costs we pay for them vs. North America ones.

### DIFF
--- a/worker-pools.yml
+++ b/worker-pools.yml
@@ -1584,9 +1584,9 @@ pools:
             maxTaskRunTime: 2592000
             enableInteractive: true
       minCapacity: 0
-      maxCapacity: 96
+      maxCapacity: 97
       implementation: generic-worker/worker-runner-linux
-      regions: [us-central1, us-west1, us-east1]
+      regions: [us-central1, us-west1, us-east1, europe-west4]
       image: monopacker-translations-worker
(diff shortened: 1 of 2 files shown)


worker-pools.yml
```diff
@@ -1584,9 +1584,9 @@ pools:
             maxTaskRunTime: 2592000
             enableInteractive: true
       minCapacity: 0
-      maxCapacity: 96
+      maxCapacity: 97
       implementation: generic-worker/worker-runner-linux
-      regions: [us-central1, us-west1, us-east1]
+      regions: [us-central1, us-west1, us-east1, europe-west4]
```
jcristau, line 1589: Should the new worker be in a dedicated pool to be able to 1) specifically target it, 2) set maxCapacity separately from the other regions?

bhearsum: Putting in a separate worker pool wouldn't make it practical to use, unfortunately. We need to try it out a real training (that uses full sized datasets) to get a good idea of how useful it is. We've got enough going on in these pools that it'll end up getting used.

(I would love to be able to set capacities by region, but I'm not sure _that's_ easily doable.)

jcristau (approved):

PR #18, opened 2024-07-11
Bug 1907217: remove `*` branch from RelEng github repositories and `maple`

bhearsum:
Stage one of the plan written out in the bug.

This _should_ be quite safe. For Maple, all of the scopes that `branch:*` has are already granted by `branch:default`.

For GitHub we have some differences, but as far as I can tell, only where the explicit branches have more scopes than `branch:*` due to grants such as https://github.com/mozilla-releng/fxci-config/blob/944ea85da779ab430e932f9829f1f02bb11ee11c/grants.yml#L1185-L1200.

Original revision/discussion in https://phabricator.services.mozilla.com/D216239

`ci-admin diff` for this is in https://gist.github.com/bhearsum/61df6555dea05bebb9c12b491d6259ce

### DIFF
--- a/projects.yml
+++ b/projects.yml
@@ -450,7 +450,6 @@ maple:
   repo_type: hg
   access: scm_level_3
   branches:
-    - name: "*"
     - name: default
   trust_domain: gecko
   features:
@@ -598,8 +597,6 @@ taskgraph:
   repo_type: git
   trust_domain: taskgraph
   branches:
-    - name: "*"
-      level: 3
     - name: main
       level: 3
   features:
@@ -617,8 +614,6 @@ mozilla-taskgraph:
   trust_domain: taskgraph
   trust_project: mozilla-taskgraph
   branches:
-    - name: "*"
-      level: 3
     - name: main
       level: 3
   features:
@@ -842,8 +837,6 @@ k8s-autoscale:
   repo: https://github.com/mozilla-releng/k8s-autoscale
   repo_type: git
   branches:
-    - name: "*"
-      level: 3
     - name: master
       level: 3
     - name: production
@@ -889,8 +882,6 @@ shipit:
     trust-domain-scopes: true
   trust_domain: releng
   branches:
-    - name: "*"
-      level: 3
     - name: main
       level: 3
     - name: production
@@ -902,8 +893,6 @@ tooltool:
   repo: https://github.com/mozilla-releng/tooltool
   repo_type: git
   branches:
-    - name: "*"
-      level: 3
     - name: master
       level: 3
     - name: production
@@ -926,8 +915,6 @@ treestatus:
   repo: https://github.com/mozilla-releng/treestatus
   repo_type: git
   branches:
-    - name: "*"
-      level: 3
     - name: master
       level: 3
     - name: production
@@ -948,8 +935,6 @@ balrog:
   repo_type: git
   trust_domain: releng
   branches:
-    - name: "*"
-      level: 3
     - name: main
       level: 3
   features:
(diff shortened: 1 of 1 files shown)


projects.yml
```diff
@@ -963,8 +948,6 @@ scriptworker:
   repo: https://github.com/mozilla-releng/scriptworker
   repo_type: git
   branches:
-    - name: "*"
-      level: 3
```
bhearsum, line 967: Bringing over a comment from @jcristau in the phab revision:
> https://github.com/mozilla-releng/scriptworker/blob/bcbdbba5399755bc43eb609773404ea34408f76a/.taskcluster.yml#L9 doesn't special-case main, so this will presumably cause scope errors if someone pushes to a non-main branch?

To which I replied:
> Good callout. I can adjust that (and any other repos with a similar configuration).

> This really gets at the heart of this change though: projects.yml (unless a * is being used) is being more strict/explicit than it has in the past. I _think_ that's a good thing in most cases. (I could perhaps see an argument for having level 1 scopes for branch:* in all cases...but I'm not eager to do that unless there's a strong reason.)

(I will be following up on this more before this lands.)

bhearsum: I went over everything being touched here. [ship it](https://github.com/mozilla-releng/shipit/commit/d63a6f05b7a8be7b65df609cf7e8c50df3e4b4dc), [scriptworker](https://github.com/mozilla-releng/scriptworker/commit/7c833f6f192bc352b66e7609ea025302df95160e), [staging-xpi-manifest](https://github.com/mozilla-releng/staging-xpi-manifest/commit/26c95d2a076dce4234dca8d50b0f98330f34a81f), [staging-adhoc-signing](https://github.com/mozilla-releng/staging-adhoc-signing/commit/ed0260a4a93f3bb5e58695de690e7576a5148285), and [staging-xpi-private](https://github.com/mozilla-releng/staging-xpi-private/commit/119cf04c8e1776508a235b8695abea237f42f4b3)'s `.taskcluster.yml` files are now all synced up with what's in projects.yml.

I also that `tooltool` has a main branch that we build for, so I added it here.

projects.yml
```diff
@@ -1148,8 +1133,6 @@ staging-xpi-manifest:
   repo: https://github.com/mozilla-releng/staging-xpi-manifest
   repo_type: git
   branches:
-    - name: "*"
-      level: 1
```
jcristau, line 1152: For the staging / level 1 repos, we may actually want to remove the `main`/`master` branch and keep just `*`?
AFAIK using the default branch is only necessary when we want to change actions, but for everything else using other branches is probably better (less chance of collision)?

bhearsum: Hm, I think that's probably reasonable. If we're doing that, we should probably keep a level 1 `*` branch in level 3 repos as well (I keep going back and forth on whether or not that's a good idea...but I don't think it does any harm).

>  AFAIK using the default branch is only necessary when we want to change actions, but for everything else using other branches is probably better (less chance of collision)?

Can you clarify what you mean by "less chance of collision"? I'm not sure I see how it connects to the rest of the comment.

jcristau: I don't think this applies to the same extent to other repos, where there's only a couple of branches we care about and the rest are pretty much only used for PRs, which get their own graphs.

For the staging repos though the entire point is to make arbitrary changes without going through PRs.  I think of those a bit like try, where each push is basically its own mercurial head, and branch names are irrelevant.

bhearsum: If you're exclusively talking about staging repos, yeah, I agree those are different. If we're talking about non-staging level 1 repos as well (of which there are many that aren't staging), I'm not sure I see those as any different from level 3 repos.

jcristau: I was thinking about the staging repos specifically, yes.  Sorry for the confusion.

bhearsum: Oh sure - yes. Let's do that.

jcristau (approved):
lgtm, thanks!

PR #20, opened 2024-07-11
ci: improve the look of notify job's Slack messages

ahal:
Some improvements to the Slack message that the notify task runs.

Unfortunately I can't access the webhook url secret anymore (and I'm afraid if I revoke it, I won't be able to re-create it). So I haven't been able to it fully.. but I simulated creating the `SLACK_MESSAGE` variable locally and it looks like it should work.


gabrielBusta (approved):

gabrielBusta:
We will see what they look like. It's just aesthetic so it seems low risk * stamps *

ahal:
Well I'm worried I may have messed up some shell scripting stuff, but I'll be prepared to do a follow-up fix or backout.

PR #30, opened 2024-07-16
fix: add retries for .taskcluster.yml fetches

bhearsum:
We get this intermittently for whatever reason, and it's constantly busting deployments.

I tried to debug the 404s a bit but didn't get anywhere; I don't think it's worth digging into more unless we see it with retries.


hneiva (approved):
LGTM

src/ciadmin/generate/tcyml.py
```diff
@@ -49,15 +49,13 @@ async def get(repo_path, repo_type="hg", revision=None, default_branch=None):
...
-                response.raise_for_status()
-                result = await response.read()
-        except aiohttp.ClientResponseError as e:
-            if e.status == 404:
-                result = None
-            else:
-                raise e
+        client = RetryClient(
+            client_session=aiohttp_session(),
+            retry_options=ExponentialRetry(attempts=5, statuses={404}),
```
jcristau, line 54: It seems odd to me that a 404 would be intermittent; are we sure that's what's happening?

We should probably retry on 5xx though, if we're adding retries...

jcristau (approved):
Intermittent 404s still seem rather wonky, but I don't have a better idea so let's go with this.

PR #31, opened 2024-07-16
bug 1907211: use europe-west4 instances in all preemptible translations GPU worker pools

bhearsum:
Our small test worked out well, and the cost is only marginally more than the US instances. Let's open up some more capacity here.

We're getting the GPU quota raised for this region in https://mozilla-hub.atlassian.net/browse/RELOPS-1011, so let's hold off merging this until that's dealt with, on the off chance we don't get the new limit we'd like.


aerickson (approved):
lgtm

PR #37, opened 2024-07-18
Bug 1908775 - Create a gecko-3/b-linux-gcp-aarch64 worker pool

gabrielBusta:
[Bug 1908775: Create a gecko-3/b-linux-gcp-aarch64 worker pool to run aarch64 tasks on level 3](https://bugzilla.mozilla.org/show_bug.cgi?id=1908775)


worker-pools.yml
```diff
@@ -250,6 +253,10 @@ pools:
       # the "t2a" machine series is only available in the "a", "b", and "f" zones of the "us-central1" region.
       regions: [us-central1]
       image: monopacker-ubuntu-2204-wayland-arm64
+      security:
+        by-chain-of-trust:
+          trusted: trusted
+          default:
```
jcristau, line 259: This seems unused AFAICT?

jcristau (changes requested):
Don't you also need a l3 images pool?

worker-pools.yml
gabrielBusta: Weird! This was failing some check as I was making these changes. But seems like I changed something else that made the error go away.

jcristau (approved):

PR #43, opened 2024-07-22
fix: add type annotations to classes with attrs usage

bhearsum:
pyright does not support attrs well, and throws tons of errors when these aren't annotated: https://github.com/microsoft/pyright/issues/146

### DIFF
--- a/src/ciadmin/generate/ciconfig/actions.py
+++ b/src/ciadmin/generate/ciconfig/actions.py
@@ -20,10 +20,10 @@ DEFAULT_INPUT_SCHEMA = {
 
 @attr.s(frozen=True)
 class Action:
-    trust_domain = attr.ib(type=str)
-    level = attr.ib(type=int)
-    action_perm = attr.ib(type=str)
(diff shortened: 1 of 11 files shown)


bhearsum:
well, forget this. pyright doesn't like attrs without type annotation, and attrs complains if you add type annotations.

jcristau:
@bhearsum should we consider replacing attrs with dataclass here?

bhearsum:
I looked into that a bit. We could, but not without some loss in functionality, per https://www.attrs.org/en/stable/why.html:
> Data Classes are intentionally less powerful than attrs. There is a long list of features that were sacrificed for the sake of simplicity and while the most obvious ones are validators, converters, [equality customization](https://www.attrs.org/en/stable/comparison.html#custom-comparison), or [extensibility](https://www.attrs.org/en/stable/extending.html) in general, it permeates throughout all APIs.

(we definitely use validators and converters)

I'm not opposed to it tbh, I'm not a huge fan of attrs, but it wouldn't necessarily be trivial.

PR #51, opened 2024-08-02
chore: drop support for Python 3.9

ahal:
Previously the Jenkins job was using Python 3.9, thus why we were running tests there. Now that we're using Github Actions to deploy, we should test against the same version that uses instead (3.11).


hneiva (approved):
awesome!
Any benefits if we switch to py12 ?

ahal:
I suppose there's no reason we couldn't, and benefits would be we get to use newer features..

I'm not going to push for that here though

PR #54, opened 2024-08-06
Replace black, flake8 and isort with ruff

ahal:
Also enable the pyupgrade set of ruff rules.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,30 +1,25 @@
-[tool.black]
+[tool.ruff]
 line-length = 88
-target-version = ["py311"]
-include = '\.(wsgi|pyi?)$'
-exclude = '''
-/(
-    \.eggs
-  | \.git
-  | \.hg
-  | \.mypy_cache
-  | \.tox
-  | \.venv
-  | \.cache
-  | \.cache_py3
-  | _build
-  | buck-out
-  | build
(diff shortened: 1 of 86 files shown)


pyproject.toml
```diff
@@ -1,30 +1,25 @@
...
+known-first-party = ["ciadmin", "build_decision"]
+known-third-party = ["taskcluster", "tcadmin"]
+
 
 [tool.isort]
-multi_line_output = 3
-include_trailing_comma = true
-force_grid_wrap = 0
-use_parentheses = true
-line_length=88
```
ahal, line 28: I think ruff does these settings by default. At least there's no equivalent to them and it still formatted our imports identically.

src/ciadmin/generate/hooks.py
```diff
@@ -23,7 +21,7 @@ async def update_resources(resources):
     resources.manage("Role=hook-id:.*")
 
     for hook in hooks:
-        hook_name = "{}/{}".format(hook.hook_group_id, hook.hook_id)
+        hook_name = f"{hook.hook_group_id}/{hook.hook_id}"
```
bhearsum, line 24: Nice to see these getting rewritten automatically!

bhearsum (approved):

PR #57, opened 2024-08-12
Bug 1912260 - revert grants for Fenix beta builds on oak

gbrownmozilla:
Reverts #56 (should no longer be needed)


bhearsum (approved):
Thanks for following up! We should switch the hg repo back to level 1 as well, so we can revert the level in `projects.yml`.

gbrownmozilla:
Yes, I'll make sure to return it to level 1 in another follow up.

PR #61, opened 2024-08-14
Remove aws provisioner references

jcristau:
These have been obsolete for about 5 years.


hneiva (approved):
extra points for removing legacy code :)

PR #68, opened 2024-08-15
fix: raise exceptions when a tcyml can't be fetched during action hook generation

bhearsum:
This used to be the case prior to adding support for multiple branches =\


bhearsum:
Depends on #69.

jcristau (approved):

PR #73, opened 2024-08-16
feat: add support for in tree actions from all configured branches

bhearsum:
I believe this change is overall a good one, but it has some possibly controversial aspects and/or downsides. I'm interested in any and all feedback or pushback. There might be some ways this can be done without some or all of the downsides.

Most notably, this change more or less requires that a `GITHUB_TOKEN` is set in the environment to run `ci-admin` successfully, otherwise you'll run into rate limits when branches are fetched from Github.

The other, less serious, downside is because we generate hooks for all globbed branches, we end up generating many, many superfluous ones for repos like https://github.com/mozilla/application-services, which have many old and now-unused release branches. One way we could improve this is to _not_ use the `branches` list to determine which branches get action hooks, but have them specified separately. This has its own downside of requiring us to maintain a separate list, and possibly not working automatically when new release branches are created.

The `ci-admin diff` for this is giant and weird because so many new hooks are being created. One thing I did for additional sanity checking was to diff the list of resource names, which showed [a whole bunch of new hooks being added](https://gist.github.com/bhearsum/885e7be7f17f63467b76ef733c5ca3c2). I'd like to diff hook bodies as well, but https://bugzilla.mozilla.org/show_bug.cgi?id=1905339 is making that difficult.

This is built on a few other PRs; only the most recent 3 commits are the relevant changes.

### DIFF
--- a/src/ciadmin/generate/branches.py
+++ b/src/ciadmin/generate/branches.py
@@ -1,0 +1,72 @@
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+from asyncio import Lock
+
+import aiohttp
+from aiohttp_retry import ExponentialRetry, RetryClient
+from tcadmin.util.sessions import aiohttp_session
+
+_cache = {}
+_lock = {}
+
+
+# TODO: support private repositories. this will most likely require querying
+# GitHub as an app.
+async def get(repo_path, repo_type="git"):
+    """Get the list of branches present in `repo_path`. Only supported for
+    public GitHub repositories."""
+    if repo_type != "git":
+        raise Exception("Branches can only be fetched for git repositories!")
+
+    if repo_path.endswith("/"):
+        repo_path = repo_path[:-1]
+    branches_url = f"https://api.github.com/repos/{repo_path}/branches"
+
+    # 100 is the maximum allowed
+    # https://docs.github.com/en/rest/branches/branches?apiVersion=2022-11-28#list-branches
+    params = {"per_page": 100}
+    headers = {}
+    if "GITHUB_TOKEN" in os.environ:
+        github_token = os.environ["GITHUB_TOKEN"]
+        headers["Authorization"] = f"Bearer {github_token}"
+
+    async with _lock.setdefault(repo_path, Lock()):
+        if repo_path in _cache:
+            return _cache[repo_path]
+
+        branches = []
+        client = RetryClient(
+            client_session=aiohttp_session(),
+            retry_options=ExponentialRetry(attempts=5),
+        )
+        while branches_url:
+            async with client.get(
+                branches_url, headers=headers, params=params
+            ) as response:
+                try:
+                    response.raise_for_status()
+                    result = await response.json()
+                    branches.extend([b["name"] for b in result])
+                    # If `link` is present in the response it will contain
+                    # pagination information. We need to examine it to see
+                    # if there are additional pages of results to fetch.
+                    # See https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api?apiVersion=2022-11-28#using-link-headers
+                    # for a full description of the responses.
+                    # This icky parsing can probably go away when we switch
+                    # to a GitHub app, as we'll likely be using a proper
+                    # client at that point.
+                    for l in response.headers.get("link", "").split(","):
+                        if 'rel="next"' in l:
+                            branches_url = l.split(">")[0].split("<")[1]
+                            break
+                    else:
+                        branches_url = None
+                except aiohttp.ClientResponseError as e:
+                    print(f"Got error when querying {branches_url}: {e}")
+                    raise e
+
+        _cache[repo_path] = branches
+        return _cache[repo_path]
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -17,6 +17,7 @@ jobs:
         environment: ["staging", "firefoxci"]
     env:
       TASKCLUSTER_ROOT_URL: ${{vars.TASKCLUSTER_ROOT_URL}}
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@v4
       - name: Set up Python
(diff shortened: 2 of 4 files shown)


bhearsum:
Update: I did some additionally sanity checking by comparing the json form with and without this change with [this script](https://gist.github.com/bhearsum/aedf174be774c79a4010df170ddd80f7). That showed that the only changes to hooks that already exist are to their descriptions, as expected:
```
~/tmp/2024-08-16 ❯ python parse.py before.json after.json
project-app-services/in-tree-action-1-generic/733b05722e's description differs
project-app-services/in-tree-action-1-release-promotion/733b05722e's description differs
project-app-services/in-tree-action-3-generic/d34f673940's description differs
project-app-services/in-tree-action-3-release-promotion/d34f673940's description differs
project-comm/in-tree-action-3-backfill/3855bf7d13's description differs
project-comm/in-tree-action-3-generic/3855bf7d13's description differs
project-comm/in-tree-action-3-merge-automation/3855bf7d13's description differs
project-comm/in-tree-action-3-purge-caches/3855bf7d13's description differs
project-comm/in-tree-action-3-release-promotion/3855bf7d13's description differs
project-comm/in-tree-action-3-retrigger-decision/3855bf7d13's description differs
project-gecko/in-tree-action-2-backfill/f526da500b's description differs
project-gecko/in-tree-action-2-generic/f526da500b's description differs
project-gecko/in-tree-action-2-googleplay/f526da500b's description differs
project-gecko/in-tree-action-2-merge-automation/f526da500b's description differs
project-gecko/in-tree-action-2-purge-caches/f526da500b's description differs
project-gecko/in-tree-action-2-release-promotion/f526da500b's description differs
project-gecko/in-tree-action-2-retrigger-decision/f526da500b's description differs
project-gecko/in-tree-action-2-scriptworker-canary/f526da500b's description differs
project-gecko/in-tree-action-3-backfill/f526da500b's description differs
project-gecko/in-tree-action-3-generic/f526da500b's description differs
project-gecko/in-tree-action-3-googleplay/f526da500b's description differs
project-gecko/in-tree-action-3-merge-automation/f526da500b's description differs
project-gecko/in-tree-action-3-purge-caches/f526da500b's description differs
project-gecko/in-tree-action-3-release-promotion/f526da500b's description differs
project-gecko/in-tree-action-3-retrigger-decision/f526da500b's description differs
project-gecko/in-tree-action-3-scriptworker-canary/f526da500b's description differs
project-mozillavpn/in-tree-action-1-generic/db45fbd59d's description differs
project-mozillavpn/in-tree-action-1-release-promotion/db45fbd59d's description differs
project-mozillavpn/in-tree-action-3-generic/db45fbd59d's description differs
project-mozillavpn/in-tree-action-3-release-promotion/db45fbd59d's description differs
project-releng/in-tree-action-3-generic/808e8c8213's description differs
project-releng/in-tree-action-3-generic/e4fd8f3800's description differs
project-releng/in-tree-action-3-generic/febcafe5a5's description differs
project-scriptworker/in-tree-action-3-generic/3cbdeb9e53's description differs
project-translations/in-tree-action-1-generic/c49d0f0c50's description differs
project-translations/in-tree-action-1-train/c49d0f0c50's description differs
```

With all of this in mind, this should be perfectly _safe_ to take, assuming the approach is desired.

jcristau (approved):
Makes sense to me.

src/ciadmin/generate/branches.py
```diff
@@ -0,0 +1,72 @@
...
+        repo_path = repo_path[:-1]
+    branches_url = f"https://api.github.com/repos/{repo_path}/branches"
+
+    # 100 is the maximum allowed
+    # https://docs.github.com/en/rest/branches/branches?apiVersion=2022-11-28#list-branches
+    params = {"per_page": 100}
+    headers = {}
+    if "GITHUB_TOKEN" in os.environ:
+        github_token = os.environ["GITHUB_TOKEN"]
+        headers["Authorization"] = f"Bearer {github_token}"
```
ahal, line 34: Which token are you planning to use here, one of the automation user accounts?

Instead, what if we authenticated using a `taskcluster-firefoxci` app installation token? This way we don't need to add to the maintenance burden around Github user accounts. This is also guaranteed to be installed on any repo that we'd need to generate branches for, plus it has higher rate limits than a regular user token.

You can use https://github.com/mozilla-releng/simple-github to authenticate easily with an app installation (you'll  need to go to https://github.com/organizations/mozilla-releng/settings/apps/firefoxci-taskcluster to generate a private key.. I can do this for you if you don't have permission).

ahal: I think it might be nice to use `simple-github` even if we don't use an app installation.. There's not currently a benefit to this other than simplifying code, but once https://github.com/mozilla-releng/simple-github/issues/20 is fixed we'll get some additional benefit.

ahal (changes requested):
I agree with the concept, but have suggestion for the implementation that feels worth pursuing.

src/ciadmin/generate/branches.py
bhearsum: The main reason I was avoiding using an app installation was because it would make local runs prohibitively difficult, or we'd need to maintain a separate (token based) code path for them (unless we're willing to make people have their own local runs, or perhaps create a shared app for everyone to use). If there's a way to avoid this problem I'm all for using an app! It certainly has the added benefit of unlocking support for private repos here.

I agree that simple-github should be used regardless of whether we use an app or not. I didn't realize it supported non-app authentication, so I didn't consider it originally.

ahal: I think supporting both would be pretty simple, the token method could be a fallback.

If we don't do the app installation, how does this work? Would we need to grant the `moz-automation` Github user access to every repository in Taskcluster? Or was there something easier you had in mind?

bhearsum: On using an app, after your comments and some reflection I agree that it's probably no big deal to use an app in GHA, and a token for local use.

On private repositories, we simply do not support hook generation for them at the moment (see https://github.com/mozilla-releng/fxci-config/blob/6816b5c1ce88f1f32cc506e7114230fc622a307c/src/ciadmin/generate/in_tree_actions.py#L53-L56). Using an app in GHA would open up that possibility, but we would still run into issues in local usage unless everyone is using a token (or app) that has access to all configured private repositories. I'm honestly not sure what is better: don't support private repos at all (until we're forced to) or only support them in GHA. The latter runs the risk of oversights or bustage because we can't easily run diffs that are equivalent to what would be deployed.

bhearsum: Oh, while I was rebasing this I realized the other reason that tokens are a bit easier: they're available automatically in GHA workflows. There's no secret creation or management to deal with. (Obviously the ones that GHA provides won't have access to private repos though.)

ahal: Thanks, this sounds fine for now then! Were you still planning to switch to simple-github? We can leave that for a follow-up if you need to land this soon.

ahal (approved):

src/ciadmin/generate/branches.py
bhearsum: I was going to, but I ended up discovering that I need access to headers for pagination, which [simple-github doesn't give me](https://github.com/mozilla-releng/simple-github/blob/2afb3503e75d789392aae7186d641d96cf7cdc17/src/simple_github/client.py#L142), so it's a bit of pain to use at the moment. Because of that, I'll leave it for a future change.

PR #85, opened 2024-09-04
fix: remove unused generic-worker config from translation docker worker pools

bhearsum:
This config is doing nothing. As far as I can tell, interactive tasks are enabled by default on docker-worker, and there is no limit on maximum run time - so the things that these settings were trying to achieve are already taken care of.


petemoore (approved):
Sweet! Thanks Ben!

jwmoss (approved):

PR #87, opened 2024-09-06
Bug 1917007 - Setup afranchuk/process-top-crashes with Taskcluster

ahal:


ahal:
This is going to fail the checks until the repo has a working Taskgraph setup.

bhearsum (approved):

ahal:
I confirmed with afranchuk that it's intentional that this is being added to a fork. There's some divergence between this fork and the main repo that will eventually be resolved, at which point we'll enable it there instead.

PR #89, opened 2024-09-09
Revert "Bug 1917007 - Enable taskgraph-cron feature for afranchuk/process-top…"

jcristau:
Reverts mozilla-releng/fxci-config#88

The cron task is erroring out every 15 minutes because the repo doesn't have a .cron.yml yet.

### DIFF
--- a/projects.yml
+++ b/projects.yml
@@ -1294,7 +1294,6 @@ process-top-crashes:
     github-pull-request:
       policy: public
     taskgraph-actions: true
-    taskgraph-cron: true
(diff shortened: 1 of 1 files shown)


bhearsum:
Maybe not necessary anymore given #90?

jcristau:
Right, we don't need this when/if https://github.com/afranchuk/process-top-crashes/pull/3 is merged or we update build-decision to include #90.

bhearsum (approved):

PR #95, opened 2024-09-11
Remove south central us region + update cacert.pem

jwmoss:
Per [bug1915855 ](https://bugzilla.mozilla.org/show_bug.cgi?id=1915855) this updates the windows worker pools to an updated cacert.pem file + removes south central us as we keep running into regional quota issues.


rcurranmoz (approved):
lgtm

PR #97, opened 2024-09-11
ci: add a task that generates the config as an artifact

ahal:


ahal:
The goal here was to upload an artifact that other tools can grab from the index. I wanted a way to capture all defined worker-pools, which the TC API doesn't provide.

Looks like I'll need to supply a GITHUB_TOKEN or we'll need to switch to an app installation though.

PR #103, opened 2024-09-18
Bug 1919579 - Remove unused mozillavpn b-win2012 pools

ahal:
These pools are no longer configured anywhere in mozilla-vpn-client


jcristau (approved):
There's some overlap with #102 FYI

cc @jwmoss

ahal:
Yeah, talked to Jonathan, let's just handle it there.

PR #104, opened 2024-09-18
refactor: use simple-github for detecting branches

ahal:
This allows us to use app auth in addition to token auth.


ahal:
I realize that this isn't really simplifying much. I'm mostly doing this because I want to stand up some automation that will require auth here, and I want to stop relying on the `moz-automation` Github accounts and start using app auth for everything instead.

ahal:
Also I filed https://github.com/mozilla-releng/simple-github/issues/49 to add implicit support for pagination in `simple-github`. It'll get there eventually :p

bhearsum (approved):
Thanks for this enhancement/follow-up. Consider updating [the GITHUB_TOKEN warning](https://github.com/mozilla-releng/fxci-config/blob/91ecb36ff8341d7def7b7ab3e88bc9c37c2e5feb/src/ciadmin/boot.py#L50-L53) to also mention app configuration. (I could go either way on this...it's not something someone is likely to use locally, but I suppose it could conceivably come up in some future CI configuration.)

ahal:
Thanks, I think app auth is complicated enough that it'll probably just add confusion. I think we should just omit it from the warning.

PR #105, opened 2024-09-18
chore: remove some unused dependencies

ahal:


hneiva (approved):
weird that removing some packages from base.in didn't remove from the compiled .txt
I'll assume those are required by other packages.

bhearsum:
> weird that removing some packages from base.in didn't remove from the compiled .txt I'll assume those are required by other packages.

Yeah, there's comments in those files to that effect: https://github.com/mozilla-releng/fxci-config/pull/105/files#diff-0940fec0afe21798f617cc4522db676f15769ae28a72992b8d19216ef7b1c3b5R822-R825

ahal:
Yeah, this is mostly a no-op in practice, but makes what we actually depend on more explicit. This happened because we basically used to re-implement the `tc-admin` cli code in this repo, but at some point many years ago we got rid of that code. That's why most of these libraries are still needed (because `tc-admin` is still using them).

PR #109, opened 2024-09-23
Relops1077: Add Azure Shared Image Gallery support 

markcor:
WIP


.worker-images.yml.swp
jwmoss, line 1: Is this required?

.worker-pools.yml.swp
jwmoss, line 1: Is this required?

markcor:
There are four items I am trying to address here:

1. Creation of Arm64 Alpha builder pool
2. Removal of the unused/no longer needed code in worker_pools.py.
3. Add support to use images from the shared image gallery. This includes the introduction of image versions.
4. Add support to include the SBOM/manifest URL in the GW config.
  
@ahal Any suggestions on the following:

How to handle the bare except that pre-commit is failing.

I want to add the SBOM/manifest URL to the GW config for Windows images. This would be under the `workerTypeMetaData` in the GW config:

```
"workerTypeMetadata": {  
  "ImageManifest": "manifest URL"
  },
```
It should probably go around here in the `get_azure_provider_config` function:

```
                "workerConfig": merge(
                    worker_config,
                ),
```
However, I am not sure on the how. Also, note that part of this will be a temporary measure until we have all Windows images hosted in a Shared Image Gallery. Then some of the conditional language can go away, and we will likely construct manifest URL instead of hardcoding it.

src/ciadmin/generate/worker_pools.py
```diff
@@ -201,109 +201,77 @@ def get_azure_provider_config(
...
+
+    launch_configs = []
+    for location in locations:
+        for vmSize in vmSizes:
+            loc = location.replace("-", "")
+            DeploymentId = image.image_id(provider_id, "deployment_id")
+            try:
+                manifest = image.image_id(provider_id, "manifest")
+            except:
+                pass
```
ahal, line 213: This `pass` will cause problems later because `manifest` (and `version`) will actually be undefined. So even referencing them will be a runtime error. Instead try doing `manifest = None`.

src/ciadmin/generate/worker_pools.py
```diff
@@ -201,109 +201,77 @@ def get_azure_provider_config(
...
+            subscription_id = f"/subscriptions/{subscription_id}"
+            resource_suffix = f"{location}-{purpose}"
+            rgroup = f"rg-{resource_suffix}"
+            vnet = f"vn-{resource_suffix}"
+            snet = f"sn-{resource_suffix}"
+            subnetId = (
+                f"{subscription_id}/resourceGroups/{rgroup}/providers/"
+                f"Microsoft.Network/virtualNetworks/{vnet}/subnets/{snet}"
+            )
+            if "version" in locals() and version:
```
ahal, line 231: Oh I see, you're checking that it's defined here. Probably still better to just set it to `None` in the `except` clause and then change this to `if version:`.

src/ciadmin/generate/worker_pools.py
```diff
@@ -201,109 +201,77 @@ def get_azure_provider_config(
             "wstServerURL", azure_config["wst_server_url"]
         )
     tags = config.get("tags", {})
-    # Temporary conditional language while we move pools over to the newer
-    # method that is in the else clause
-    # Refrence https://mozilla-hub.atlassian.net/browse/RELOPS-212
-    if config.get("old-style-worker-config"):
```
ahal, line 207: Thanks for deleting this! Though it would be nice to remove this in a separate PR (or at least a separate commit) to make the diff easier to read.

ahal (changes requested):

src/ciadmin/generate/worker_pools.py
markcor: Good point. I am going to separate this into different PRs.

PR #111, opened 2024-09-24
Allow build-decision to get files from private github repositories

Eijebong:
The private repository stuff was already implemented in the path to get push info but not in the `get_file` method.

### DIFF
--- a/build-decision/src/build_decision/repository.py
+++ b/build-decision/src/build_decision/repository.py
@@ -36,18 +36,25 @@ class Repository:
 
         If the file is not found, this returns None.
         """
+        headers = {}
+
         if self.repository_type == "hg":
             if revision is None:
                 revision = "default"
             url = f"{self.repo_url}/raw-file/{revision}/{path}"
         elif self.repository_type == "git":
             repo_url = self.repo_url
-            if revision is None:
-                revision = "master"  # FIXME: Use api to get default branch
+
+            ref_param = ""
+            # If no ref is given, the github API will default to the default branch
+            if revision is not None:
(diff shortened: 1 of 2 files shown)


build-decision/src/build_decision/repository.py
```diff
@@ -45,9 +47,9 @@ def get_file(self, path, *, revision=None):
             if revision is None:
                 revision = "master"  # FIXME: Use api to get default branch
             if repo_url.startswith("https://github.com/"):
-                if repo_url.endswith("/"):
-                    repo_url = repo_url[:-1]
-                url = f"{repo_url}/raw/{revision}/{path}"
+                url = f"https://raw.githubusercontent.com/{self.repo_path}/{revision}/{path}"
```
jcristau, line 50: Should we use the documented API (https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28) here instead of the githubusercontent domain?

Eijebong: Yeah probably, I didn't think about checking if they had an API. I'll test it and change it if it works

jcristau: As far as I can tell you can get the raw contents with `Accept: application/vnd.github.raw+json`?

Eijebong: You're right... I only looked at their examples which doesn't list the raw one... I'll change it

jcristau (approved):
lgtm, thanks!

We could probably also handle the `git@github.com:...` case this way, if/when that turns out to be needed.

PR #115, opened 2024-09-25
Relops 1077 - removed old-style-worker-config from get_azure_provider…

markcor:
…_config

Cleaning up the `get_azure_provider_config`. Removing the unused `old-style-worker-config`. This is preparation to add additional functionality.


ahal (approved):
Thanks!

PR #116, opened 2024-09-26
Update thunderbird-releng group to use thunderbird_releng LDAP.

jfx2006:
Bug 1921006 updates the thunderbird_releng LDAP group to current staff.


coreycb (approved):
Looks good to me, thanks Rob.

PR #118, opened 2024-09-26
ci: ensure apply jobs continue even if the other fails

ahal:
During a recent deploy, we realized that the apply jobs will get cancelled if the other one fails.

This is bad because the apply is not atomic, so if it gets cancelled, it could leave the cluster in a bad state.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -42,6 +42,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: check
     environment: apply-${{matrix.environment}}
+    continue-on-error: true
     strategy:
       matrix:
         environment: ["staging", "firefoxci"]


jcristau (commented):
I'm not super clear on what this does, but it sound like it would end up marking the workflow as successful in case of an error?
Should we disable `jobs.apply.strategy.fail-fast` instead maybe?

bhearsum:
> I'm not super clear on what this does, but it sound like it would end up marking the workflow as successful in case of an error? Should we disable `jobs.apply.strategy.fail-fast` instead maybe?

I haven't tested it, but based on https://www.kenmuse.com/blog/how-to-handle-step-and-job-errors-in-github-actions/ it sounds like the workflow (and even the individual job) will still fail.

ahal:
To be frank, I'm not 100% sure either. I think this will do what we want (mark the job as `fail` but don't abort the other "matrices"). Github Actions still confuses me a lot though, so I could be misreading the docs.

ahal:
I think one source of confusion is that this flag has completely different behaviour depending on whether it's set on the top-level job, or on an individual step.

Here's the docs for when you set it on a job (like I am in this patch):
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

The example they give below looks to be the same use case we have.

jcristau (approved):

PR #119, opened 2024-09-26
Relops1077 - Add support ...

markcor:
... to use versioned Azure Images stored in Shared Image Galleries.
Add ronin_b_windows11_a64_24h2_builder_alpha image alias. This includes a `SBOM` field. This field is in place for future functionality to have it included in Win workers' GW config. 
Add win11-a64-24h2-builder-alpha pool.


jwmoss (approved):

src/ciadmin/generate/worker_pools.py
```diff
@@ -207,7 +207,11 @@ def get_azure_provider_config(
         for vmSize in vmSizes:
             loc = location.replace("-", "")
             DeploymentId = image.image_id(provider_id, "deployment_id")
-            ImageId = image.image_id(provider_id, location)
+            try:
+                version = image.image_id(provider_id, "version")
+            except Exception:
+                version = None
+            DeploymentId = image.image_id(provider_id, "deployment_id")
```
bhearsum, line 214: It looks like there's now two copies of this line? One of them probably ought to be removed.

src/ciadmin/generate/worker_pools.py
```diff
@@ -207,7 +207,11 @@ def get_azure_provider_config(
         for vmSize in vmSizes:
             loc = location.replace("-", "")
             DeploymentId = image.image_id(provider_id, "deployment_id")
-            ImageId = image.image_id(provider_id, location)
+            try:
+                version = image.image_id(provider_id, "version")
+            except Exception:
```
bhearsum, line 212: It looks like it might just be `KeyError` that you want to catch here.

src/ciadmin/generate/worker_pools.py
```diff
@@ -221,10 +225,18 @@ def get_azure_provider_config(
...
                 f"Microsoft.Network/virtualNetworks/{vnet}/subnets/{snet}"
             )
-            imageReference_id = (
-                f"{subscription_id}/resourceGroups/{image_rgroup}/providers/"
-                f"Microsoft.Compute/images/{ImageId}-{DeploymentId}"
-            )
+            if version is not None:
+                ImageId = image.image_id(provider_id, "name")
+                imageReference_id = (
+                    f"{subscription_id}/resourceGroups/rg-packer-worker-images/providers/"
```
bhearsum, line 231: What is `rg-packer-worker-images`? Does it need to be hardcoded here? At the very least this should probably be raised to a constant at the top of the file rather than buried here. Preferably, `image_rgroup` would be correct both with and without a `version`. It looks like that's already in the yml, even?

src/ciadmin/generate/worker_pools.py
markcor: Do you have any suggestions on how to catch here? 

Side note: `tc-admin generate --environment=firefoxci` ran locally without error.

bhearsum: `except KeyError` should do the trick.

The main concern about catching all exceptions here is that you might end up catching some _other_ error that doesn't indicate you should take the `except` branch.

markcor:
Took a different approach. Updated all the image aliases to include a version field. 

As well as added a resource group field to aliases. It makes more sense that this value is connected to the image, not the pool. The pool image resource group field should be removed in a future PR.

tests/ciadmin/test_generate_worker_pools.py
```diff
@@ -212,6 +218,17 @@ def assert_guest_accelerators(pool):
...
+        pytest.param(
+            "azure",
+            {
+                "storageProfile": {
+                    "imageReference": {
+                        "id": "/subscriptions/108d46d5-fe9b-4850-9a7d-8c914aa6c1f0/rgroup_id/rg-packer-worker-images/providers/Microsoft.Compute/galleries/name_id/images/name_id/versions/ver_id"
+                    }
+                },
+            },
+            id="azure_new_format",
```
ahal, line 230: You need to create an assertion function that matches this `id`. Namely:

```
def assert_azure_new_format(pool):
    # put whatever assertions you want here
```

ahal (changes requested):

tests/ciadmin/test_generate_worker_pools.py
```diff
@@ -212,6 +218,17 @@ def assert_guest_accelerators(pool):
             id="guest_accelerators",
         ),
         pytest.param("azure", None, id="azure_basic"),
+        pytest.param(
+            "azure",
+            {
+                "storageProfile": {
+                    "imageReference": {
+                        "id": "/subscriptions/108d46d5-fe9b-4850-9a7d-8c914aa6c1f0/rgroup_id/rg-packer-worker-images/providers/Microsoft.Compute/galleries/name_id/images/name_id/versions/ver_id"
```
ahal, line 226: I suspect you don't want to put this config here. This `pytest.param` is the inputs to the pool generation, so I *think* this is the output you want to verify gets generated correct?

Assuming that's the case, you'll want to assert the result is equivalent to  this in your assertion function (see prior comment).

jwmoss (approved):

bhearsum (approved):
Sorry for the delay in reviewing this. I'm not well equipped to review the particulars of the pool configuration (I think @jwmoss has covered that part though?), but the new code and tests look fine to me.

ahal (approved):

PR #125, opened 2024-09-30
Grant snap_sudo os group on t-linux-2404-wayland

jwmoss:


aerickson (approved):

jcristau (changes requested):
I don't think wildcards in the middle of a scope work, they need to be at the end (https://docs.taskcluster.net/docs/reference/platform/auth/scopes#stars-in-scopes)

jwmoss:
> I don't think wildcards in the middle of a scope work, they need to be at the end (https://docs.taskcluster.net/docs/reference/platform/auth/scopes#stars-in-scopes)

Mind checking again?

jcristau (approved):

PR #126, opened 2024-10-03
Bug 1906337 - Remove experimental 'dagger' pool

ahal:


ahal:
I never found the time to get this working, and I don't think it's worth pursuing at the moment. Dagger is basically a `docker-worker` replacement for us. I think the main benefit here would be the ability to re-use community pipelines, but tbh that doesn't seem super compelling. A nice to have, but there are more important things to work towards.

jcristau (approved):

PR #127, opened 2024-10-03
Bug 1922252 - Remove l10n-cross-channel from comm-central cron targets.

jfx2006:


jfx2006:
Job is removed from comm-central .cron.yml.

jcristau (approved):

PR #132, opened 2024-10-11
add test image for translations worker with dnsmasq instead of system-resolved

bhearsum:
Related to https://github.com/mozilla-platform-ops/monopacker/pull/149 and https://github.com/mozilla/firefox-translations-training/issues/549


aerickson (approved):
lgtm

PR #134, opened 2024-10-15
Create x11 24.04 image for testing

jwmoss:


worker-pools.yml
```diff
@@ -1581,6 +1581,31 @@ pools:
             - <<: *persistent-disk
               diskSizeGb: 60
           machine_type: c2-standard-4
+  - pool_id: '{pool-group}/t-linux-2404-x11-relsre'
+    description: Workers for x11 on Ubuntu 24.04. Used by RelSRE to qualify images.
+    owner: jmoss@mozilla.com
+    email_on_error: true
```
markcor, line 1587: We should move this to a google group.

markcor (approved):

worker-pools.yml
jwmoss: Yeah I think that was for testing. I'll update it.

jcristau (changes requested):
I have questions about this.  I'd like to understand what the use case is for such an image/pool, because in my mind this isn't wanted/necessary.

PR #148, opened 2024-10-23
fix: create mozilla-3/decision-gcp pool

ahal:
This is needed for `mozilla/fx-desktop-qa-automation` which just transitioned to a level 3 project.

An argument could be made that they should instead get their own trust domain, but personally I feel like *all* projects should be L3. Distinguishing scopes between PRs and pushes is extremely common, even in projects that aren't particularly security sensitive.

Where I think we should draw the line for the `mozilla` trust domain, is if projects start needing scriptworkers for some reason.

### DIFF
--- a/worker-pools.yml
+++ b/worker-pools.yml
@@ -274,6 +274,8 @@ pools:
       - pool-group: mobile-3
         chain-of-trust: trusted
       - pool-group: mozilla-1
+      - pool-group: mozilla-3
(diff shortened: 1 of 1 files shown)


bhearsum (approved):
It's difficult to think of any downside to having level 3 workers available. I think I'm on board with having scriptworkers be a line in the sand (seeing as that implies the project has some much more privileged secrets).

PR #155, opened 2024-10-30
Run os-integration tests

ahal:


taskcluster/scripts/schedule-tasks.py
```diff
@@ -0,0 +1,54 @@
...
+    assert task_graph
+
+    for label, task in task_graph.items():
+        assert isinstance(task, dict)
+        if task.get("attributes", {}).get("unittest_variant") != "os-integration":
+            continue
+
+        task_def = task["task"]
+        task_def["priority"] = "low"
+        del task_def["routes"]
```
jcristau, line 48: Presumably you need to get rid of dependencies here too?
But then need a way to handle things like docker images and fetches, that reference other task ids that would not exist on staging?

taskcluster/scripts/tc-admin-apply-staging.py
```diff
@@ -0,0 +1,30 @@
...
+auth = fetch_secret("project/releng/fxci-config/taskcluster-stage-client")
+env = os.environ.copy()
+env.update(
+    {
+        "TASKCLUSTER_ROOT_URL": "https://stage.taskcluster.nonprod.cloudops.mozgcp.net",
+        "TASKCLUSTER_CLIENT_ID": auth["clientId"],
+        "TASKCLUSTER_ACCESS_TOKEN": auth["accessToken"],
+    }
+)
+subprocess.run(["tc-admin", "apply", "--environment=staging"], env=env)
```
jcristau, line 30: check=True?

jcristau (commented):
I know this is still draft, but I was wondering how you were handling some stuff so I had a sneak peek anyway.

taskcluster/scripts/schedule-tasks.py
ahal: Heh, yeah good call.

My rough idea was to re-write the `MOZ_FETCHES` environment variable to use `static-url` instead. But haven't looked too closely at that yet.

bhearsum (approved):
> taskcluster/scripts/fetch-task-definition.yml

Empty file? :laughing:

PR #158, opened 2024-10-31
chore: rename production translations repository

bhearsum:
This rename is happening in https://github.com/mozilla/firefox-translations-training/pull/914. We plan to land that change and this one around the same time rather than to support both at the same time.

### DIFF
--- a/grants.yml
+++ b/grants.yml
@@ -2902,7 +2902,7 @@
   to:
     - project:
         alias:
-          - firefox-translations-training
+          - translations
         job:
           - branch:*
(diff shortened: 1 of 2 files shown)


bhearsum:
I recommend that we wait until https://github.com/mozilla/firefox-translations-training/pull/914 is landed and that repository is renamed, and then we can rebase this patch to ensure that tests patch, and land it. This should be coordinated with the rename, as Taskcluster will break in that repository after that happens until this is rolled out.

ahal (approved):

PR #166, opened 2024-11-07
Bug 1919464 - bump code-coverage/bot-gcp instance type to c2-standard-16

jcristau:
It looks like we're still OOMing on c2-standard-8.

cc @marco-c


marco-c (approved):

ahal (approved):

marco-c:
Is this applied too? So I can test if the hook now works

PR #167, opened 2024-11-07
Bug 1929673 - Grant fxci-config access to a secret containing the sta…

ahal:
…ging instance client


ahal:
This for #155 but needs to be applied before I can proceed further with that PR.

hneiva (approved):

PR #179, opened 2024-11-20
Add SBOM url to generic-worker metadata

ahal:


src/ciadmin/generate/worker_pools.py
```diff
@@ -195,10 +195,25 @@ def get_azure_provider_config(
...
-        )
+        gw_config.setdefault("wstServerURL", azure_config["wst_server_url"])
+
+    # Populate some generic-worker metadata.
+    metadata = {}
+    has_sbom = image.get(provider_id, "sbom") in (None, True)  # defaults None to True
+    if has_sbom and "sbom_url_tmpl" in azure_config:
+        context = image.clouds[provider_id].copy()
+
+        # TODO Ensure the sbom urls match defined names, then remove this.
```
bhearsum, line 208: Is this a TODO that you need to address before landing?

worker-images.yml
```diff
@@ -190,76 +194,68 @@ ronin_t_windows10_64_2009_prod:
...
-    SBOM: https://github.com/mozilla-platform-ops/worker-images/blob/main/config/win11-a64-24h2-builder-alpha.md
     name: win11_a64_24h2_builder_alpha
 ronin_b1_windows11_a64_24h2_builder:
   azure2:
     version: 1.0.0
     resource_group: rg-packer-worker-images
     deployment_id: a03478d
-    SBOM: https://github.com/mozilla-platform-ops/worker-images/blob/main/config/win11-a64-24h2-builder.md
     name: win11_a64_24h2_builder
+    sbom: false
```
bhearsum, line 217: I see that this, and at least one other entry, appear to be losing SBOMs? That seems less than ideal? (Maybe there's a good reason though?)

bhearsum (approved):
> You can see this working here: https://stage.taskcluster.nonprod.cloudops.mozgcp.net/tasks/Fm7BV5OXRG-FPG0uR4Mh7A/runs/0/logs/public/logs/live.log#L24
> 
> I wonder if it would be worth pulling in the entire JSON of the SBOM into the metadata..

I'm struggling to articulate why...but it seems like we ought to keep the link to the repo. I don't see any harm in additionally embedding the SBOM in the metadata though. I could probably be convinced to just do it instead, until we have a compelling reason to do both.

A couple of questions below, but whether or not they need any changes is up to you.

src/ciadmin/generate/worker_pools.py
ahal: No, the names as defined in `worker-images.yml` use underscore, while the urls use dashes. So the comment was more about making those two things match each other. I'll make it a bit more clear.

worker-images.yml
ahal: The old `SBOM` entries were added manually and were unused, so they were essentially just a comment. Now the sboms are moving to a new naming scheme. For whatever reason the url for this one was a 404, so I disabled it for now. I suspect jmoss just needs to tweak something in worker-images, but this way the PR can land and he can re-enable it when it's ready.

bhearsum (approved):

PR #180, opened 2024-11-20
feat(tc-admin): Add --resources parameter to tc-admin

hneiva:
Handles new CLI parameter --resources
Tries to handle --help as best as we can without interferring how tc-admin is handling arguments.


ahal (approved):
:see_no_evil:

PR #181, opened 2024-11-21
Consolidate Windows worker-pool configuration

ahal:
This makes heavier use of variants to reduce duplication in the Windows configs.


ahal:
This pull request results in an empty `tc-admin diff`.

hneiva (approved):
LGTM

PR #185, opened 2024-11-21
use new ubuntu 24.04 headless images for d2g translations images

bhearsum:
The existing ones are not used in production, so there's no real risk to doing this.


aerickson (approved):
lgtm

PR #186, opened 2024-11-21
Disable Generic Worker interactive and D2G features by default

petemoore:
Disable features unless explicitly enabled:
* Interactive
* D2G

Worker pools have can opt in to these features, but without explicitly enabling, they will be disabled.

### DIFF
--- a/worker-pools.yml
+++ b/worker-pools.yml
@@ -60,6 +60,8 @@ worker-defaults:
             disableReboots: false
             downloadsDir: Z:\downloads
             ed25519SigningKeyLocation: C:\generic-worker\ed25519-private.key
+            enableD2G: false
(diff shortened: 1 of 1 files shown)


worker-pools.yml
```diff
@@ -60,6 +60,8 @@ worker-defaults:
             disableReboots: false
             downloadsDir: Z:\downloads
             ed25519SigningKeyLocation: C:\generic-worker\ed25519-private.key
+            enableD2G: false
```
bhearsum, line 63: I don't think this is doing what you think it is. This is the default config for `generic-worker/windows`. At the very least you'll want this in the other sections here, and perhaps to remove `enableD2G` from this one?

PR #189, opened 2024-11-25
fix: clean up integration task definitions

ahal:
These tasks' definitions get completely overwritten in the `schedule_integration` transforms. The 'dependencies' and 'run' keys here were left over from a previous implementation of the feature that I ended up moving away from.


jcristau (approved):

ahal:
Realized I could also remove all the task-defaults and scopes.

PR #190, opened 2024-11-26
ci: support docker-worker tasks as well

ahal:


taskcluster/kinds/integration-image/kind.yml
```diff
@@ -0,0 +1,29 @@
...
+  worker:
+    docker-image: {in-tree: python3.11}
+    env:
+      MOZ_ARTIFACT_DIR: /builds/worker/artifacts
+    max-run-time: 3600
+  run:
+    using: run-task
+    cwd: '{checkout}'
+    command: >-
+      curl -o $MOZ_ARTIFACT_DIR $FETCH_URL
```
jcristau, line 23: `--fail --location --retry 5` (assuming we can't / don't want to use fetch-content here...)

ahal: Thanks, this was a placeholder while I was working out the dependencies. But you saved me some Googling, so appreciated! :)

The problem with `fetch-content` was that it requires knowing the hash and size, which I guess we could find a way to obtain if we were motivated enough.. but imo meh.

ahal: Although I guess I could do the same use `task-artifact` fetches and fake the TASKCLUSTER_ROOT_URL trick I'm doing with the test tasks...

ahal:
This is ready for review. I have plans to use the `firefoxci-artifact` kind to support private artifacts, but I'll do that in a follow-up PR.

taskcluster/scripts/tc-admin-apply-staging.py
```diff
@@ -13,20 +13,29 @@
...
+
 env = os.environ.copy()
 del env["TASKCLUSTER_PROXY_URL"]  # ensure we don't use the proxy client
 env.update(
     {
         "TASKCLUSTER_ROOT_URL": ROOT_URL,
         "TASKCLUSTER_CLIENT_ID": CLIENT_ID,
         "TASKCLUSTER_ACCESS_TOKEN": token,
+        "GITHUB_APP_ID": str(github_app_creds["app_id"]),
+        "GITHUB_APP_PRIVKEY": github_app_creds["ssh_privkey"],
```
bhearsum, line 38: Is there any chance of this leaking through the env? (Do we even care for staging?

bhearsum (approved):

taskcluster/fxci_config_taskgraph/transforms/integration_test.py
```diff
@@ -75,18 +32,23 @@ def patch_root_url(task_def):
...
             command[1] = [
                 "bash",
                 "-c",
                 f"TASKCLUSTER_ROOT_URL={FIREFOXCI_ROOT_URL} {' '.join(command[1])}",
             ]
+        else:
+            command[:] = [
+                "bash",
+                "-c",
+                f"TASKCLUSTER_ROOT_URL={FIREFOXCI_ROOT_URL} {' '.join(command)}",
```
jcristau, line 50: `shlex.join(command)`?

jcristau (commented):
[Is this starting to get to the point where it could do with some tests for the transforms?]

taskcluster/fxci_config_taskgraph/transforms/firefoxci_artifact.py
```diff
@@ -0,0 +1,63 @@
...
+transforms = TransformSequence()
+
+
+def get_artifact_urls(task_id: str, fetch_artifacts: list[str]) -> str:
+    queue = get_taskcluster_client("queue")
+    artifacts = queue.listLatestArtifacts(task_id)["artifacts"]
+
+    urls = [
+        queue.buildUrl("getLatestArtifact", task_id, artifact["name"]).replace(
+            "%2F", "/"
```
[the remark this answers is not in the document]
ahal: My memory is a bit fuzzy, but I think it was because we need to extract the final path segment to get the file name to download to. I suppose I could have split the url on `%2F` instead, but it seemed cleaner to normalize it.

src/ciadmin/boot.py
```diff
@@ -51,7 +51,9 @@
 
 
 def boot():
-    if not os.environ.get("GITHUB_TOKEN"):
+    if not os.environ.get("GITHUB_TOKEN") or not (
```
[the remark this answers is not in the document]
ahal: Oops, I had this fixed locally but must have forgotten to push.

ahal:
(Tasks are failing because I had previously edited the role manually, but they got overwritten)

PR #192, opened 2024-11-28
wip: make it possible to schedule translations pipeline runs

bhearsum:


bhearsum:
This is ultimately blocked on https://github.com/mozilla-services/cloudops-infra/pull/6127; I may hack around that for the moment though.

bhearsum:
After some discussion with others, this approach seems less preferable thank the one in #210, in part because of some of the hacks it requires (like https://github.com/taskcluster/taskgraph/pull/611), but also because it adds an entirely separate implementation for how to run these integration tests.

PR #200, opened 2024-12-04
ci: support private artifacts in fxci integration tests

ahal:


ahal:
Sweet, looks like this is finally working!

Some things on the TODO list:
* tests
* using cached tasks for the `firefoxci-artifact` kind

But these are things that could be done in a follow-up, so I'll flag for review now and incoporate those later (either in this PR or another).

taskcluster/scripts/fetch-firefoxci-artifacts.py
```diff
@@ -0,0 +1,53 @@
...
+
+secrets = taskcluster.Secrets(options)
+
+result = secrets.get("project/releng/fxci-config/firefoxci-artifact-client")
+assert isinstance(result, dict)
+
+credentials = result["secret"]
+assert isinstance(credentials, dict)
+
+options = {"rootUrl": FIREFOXCI_ROOT_URL, "credentials": credentials}
```
hneiva, line 27: is this intended behavior? `options` is already set up above with `rootUrl`, and used to fetch credentials, and this one for the queue. Is there a use case where they would be using different `rootUrl`?

taskcluster/scripts/fetch-firefoxci-artifacts.py
```diff
@@ -0,0 +1,53 @@
...
+queue = taskcluster.Queue(options)
+
+artifact_dir = os.environ["MOZ_ARTIFACT_DIR"]
+os.chdir(artifact_dir)
+
+task_id = os.environ["FETCH_FIREFOXCI_TASK_ID"]
+artifacts = json.loads(os.environ["FETCH_FIREFOXCI_ARTIFACTS"])
+for artifact in artifacts:
```
hneiva, line 35: could we async download the artifacts? (maybe an improvement for another PR)

hneiva (approved):
couple of minor questions, but LGTM otherwise

taskcluster/scripts/fetch-firefoxci-artifacts.py
bhearsum: AIUI, it first talks to staging to fetch the secret that contains the credentials for the Firefox CI cluster.

This might be a bit less confusing if there were two distinct `options` and/or a comment about the above.

bhearsum (changes requested):
I'm requesting changes for the hardcoded artifact name, which I assume can be adjusted. If I'm missing something there I'll happily switch this to Approved.

taskcluster/scripts/fetch-firefoxci-artifacts.py
ahal: Good idea, though it's probably pretty rare that we'd be fetching more than one artifact from a task anyway. Though thinking about it, we'd need to depend on aiohttp just for this, so I think I'll stick with synchronous for now.

taskcluster/scripts/fetch-firefoxci-artifacts.py
ahal: I'll add a comment.

ahal:
I refactored the script to use a couple functions and added an assertion to ensure we're only running this script from the staging instance. Hope that makes it more clear!

bhearsum (approved):
That's a nice refactor of the script; much clearer now!

PR #201, opened 2024-12-04
ci: ignore android-hw and macosx integration tests

ahal:
Due to an oversight on the Gecko side, we accidentally scheduled a Bitbar test. This will be fixed the next time the cron task runs, but this acts as an extra safeguard.

### DIFF
--- a/taskcluster/fxci_config_taskgraph/util/integration.py
+++ b/taskcluster/fxci_config_taskgraph/util/integration.py
@@ -41,7 +41,11 @@ def find_tasks(decision_index_path: str) -> list[dict[str, Any]]:
     tasks = []
     for task in task_graph.values():
         assert isinstance(task, dict)
-        if task.get("attributes", {}).get("unittest_variant") != "os-integration":
+        attributes = task.get("attributes", {})
+        if attributes.get("unittest_variant") != "os-integration":
+            continue
+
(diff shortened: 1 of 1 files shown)


taskcluster/fxci_config_taskgraph/util/integration.py
```diff
@@ -41,7 +41,11 @@ def find_tasks(decision_index_path: str) -> list[dict[str, Any]]:
     tasks = []
     for task in task_graph.values():
         assert isinstance(task, dict)
-        if task.get("attributes", {}).get("unittest_variant") != "os-integration":
+        attributes = task.get("attributes", {})
+        if attributes.get("unittest_variant") != "os-integration":
+            continue
+
+        if attributes.get("test_platform", "").startswith(("android-hw", "macosx")):
```
bhearsum, line 48: If we're not going to run these tasks as part of these tests, should we stop running as part of the cron task in Gecko rather than ignoring them here?

ahal: Yeah, I fixed it there already too. But the cron task won't run again until Monday, so this unblocks us in the meantime (and provides an extra safeguard from future mistakes)

bhearsum (approved):

PR #202, opened 2024-12-04
Add additional worker pools for win11-64-24H2

jwmoss:


bhearsum (commented):
Is there some context (eg: a bug or jira ticket) for this?

bhearsum (approved):

PR #208, opened 2024-12-09
ci: grant fxci-config access to secret for fetching fxci artifacts

ahal:


ahal:
This secret contains credentials for a Taskcluster client on the firefoxci instance.

ahal:
This is a follow-up to #206, I should have known I would need to grant it and included this in that PR, but forgot.

bhearsum (approved):

PR #210, opened 2024-12-10
feat: translations integration tests

bhearsum:
This adds support for running the translations training pipeline as part of integration tests. The first two commits, which add support for fetching upstream tasks that didn't run as part of a decision task and support for rewriting dependencies to the yet-to-be-scheduled tasks are the interesting parts. The other few commits are pretty boring, and simply fix a minor bug and enable the translations tests.

Note that these depends on https://github.com/mozilla-releng/fxci-config/pull/228, a yet to be opened PR in https://github.com/mozilla/translations to add a pipeline cronjob. It's currently running against the staging repository and working well though.

PR #218, opened 2024-12-12
silly change to test possible issue with taskcluster upgrade

bhearsum:
I'm seeing integration tests fail to schedule and/or find docker images in https://github.com/mozilla-releng/fxci-config/pull/210 after pulling in a recent taskgraph upgrade. I'm going to try repro'ing here to disentangle what's at fault.


bhearsum:
alright, everything scheduled here correctly...the problem is caused by something in the other PR!

PR #219, opened 2024-12-12
ci(integration): add tests for integration related transforms and scr…

ahal:
…ipts

I came up with this hare-brained idea to roll my own parameterization scheme. The inputs to the test cases go in the docstring. Then a fixture handles all the business logic and returns the result.

I really like how this turned out. It has all the benefits of using `pytest.mark.parametrize`, without the pitfalls. Namely, there's no need for dangling assertion functions that live separately from the test inputs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,3 +19,6 @@ ignore = [
 [tool.ruff.lint.isort]
 known-first-party = ["ciadmin", "build_decision", "fxci_config_taskgraph"]
 known-third-party = ["taskcluster", "tcadmin", "taskgraph"]
+
+[tool.pytest.ini_options]
+testpaths = ["tests", "taskcluster/test"]
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -484,7 +484,7 @@ click==8.1.7 \
     --hash=sha256:ae74fb96c20a0277a1d615f1e4d73c8414f5a98db8b799a7931d1582f3390c28 \
     --hash=sha256:ca9853ad459e787e2192211578cc907e7594e294c7ccc834310722b41b9ca6de
     # via tc-admin
-colorama==0.4.6 ; sys_platform == 'win32' or platform_system == 'Windows' \
+colorama==0.4.6 ; platform_system == 'Windows' or sys_platform == 'win32' \
     --hash=sha256:08695f5cb7ed6e0531a20572697297273c47b8cae5a63ffc6d6ed5c201be6e44 \
     --hash=sha256:4f1d9991f5acc0ca119f9d443620b77f9d6b33703e51011c16baf57afb285fc6
     # via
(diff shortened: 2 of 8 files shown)


bhearsum (commented):
I definitely agree that this is better than the separate assertion functions that we have in taskgraph's tests. I think I'm not really sure why we can't just use pytest better though? It looks to me like you could specify the expected result in the params as well instead of using separate functions? Eg:
```
def run_test(task_overrides, expected_task):
  # this stays the same; may also have to pass in other fixtures though?

@pytest.mark.parametrize(
  "task_overrides,expected_task",
  @pytest.mark.param(
    {
      "attributes": {"foo": "bar"}
    },
    None,
    id="missing_unittest_variant",
  ),
  pytest.param(
    {
      "attributes": {"unittest_variant": "os-integration"},
    },
    {
      ....
    },
    id="basic",
  )
)
def test_integration_test_transforms(task_overrides, expected_task):
  task = run_test(task_overrides, expected_task)
  assert task == expected_task
```

As written above, this _does_ have the downside of needing to specify the full expected task in each result. This could be improved by, eg: walking `expected_task` and only asserting the values of the leaf keys?

I don't want to say I'm a hard no on what you have here, as I _think_ the bodies of the test functions are a bit cleaner than they would be with my suggestion, but I do shudder a bit at having things that impact test outcome in comments.

I wonder if another way to go about this is to use a class for each test instead of a function, and put the overrides in class attributes?

ahal:
> As written above, this does have the downside of needing to specify the full expected task in each result. This could be improved by, eg: walking expected_task and only asserting the values of the leaf keys?

Yeah, this is the main reason I'm souring on parametrization in general. After reading that test contravariance article you linked, it made me realize that asserting the whole output leads to fragile tests (small change in code leading to large change in tests). So I really wanted to only assert what the actual test was supposed to be checking. Your suggestion about only asserting leaf nodes could work in this case, but:

1. I'm trying to find a structure that works in general (what if we want to check for an exception or stdout?)
2. When you start adding too many variables to the parametrization I find it gets very hard to read, largely due to `pytest.param` not accepting kwargs for them.

But I think I have a better solution involving rewriting the `result` fixture into a `run_test` fixture. I'll update the PR shortly.

bhearsum (approved):
> > As written above, this does have the downside of needing to specify the full expected task in each result. This could be improved by, eg: walking expected_task and only asserting the values of the leaf keys?
> 
> Yeah, this is the main reason I'm souring on parametrization in general. After reading that test contravariance article you linked, it made me realize that asserting the whole output leads to fragile tests (small change in code leading to large change in tests). So I really wanted to only assert what the actual test was supposed to be checking. Your suggestion about only asserting leaf nodes could work in this case, but:
> 
>     1. I'm trying to find a structure that works in general (what if we want to check for an exception or stdout?)
> 

Good point; this is obviously worthwhile to pursue!

>     2. When you start adding too many variables to the parametrization I find it gets very hard to read, largely due to `pytest.param` not accepting kwargs for them.

:nod:

This new approach looks like it has all the upsides and none of the downsides!

PR #221, opened 2024-12-13
fix: default value for MOZ_FETCHES in integration test transform should be an empty object

bhearsum:
An empty string is not valid JSON:
>>> import json
>>> json.loads("")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/bhearsum/.pyenv/versions/3.11.6/lib/python3.11/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bhearsum/.pyenv/versions/3.11.6/lib/python3.11/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bhearsum/.pyenv/versions/3.11.6/lib/python3.11/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

### DIFF
--- a/taskcluster/fxci_config_taskgraph/transforms/firefoxci_artifact.py
+++ b/taskcluster/fxci_config_taskgraph/transforms/firefoxci_artifact.py
@@ -42,7 +42,7 @@ def make_firefoxci_artifact_tasks(config, tasks):
                 # Add private artifacts
                 if "MOZ_FETCHES" in task_def["payload"].get("env", {}):
                     fetches = json.loads(
-                        task_def["payload"].get("env", {}).get("MOZ_FETCHES", {})
+                        task_def["payload"].get("env", {}).get("MOZ_FETCHES", "{}")
                     )
                     for fetch in fetches:
                         if fetch["artifact"].startswith("public"):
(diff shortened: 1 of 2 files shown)


taskcluster/fxci_config_taskgraph/transforms/firefoxci_artifact.py
```diff
@@ -42,7 +42,7 @@ def make_firefoxci_artifact_tasks(config, tasks):
                 # Add private artifacts
                 if "MOZ_FETCHES" in task_def["payload"].get("env", {}):
                     fetches = json.loads(
-                        task_def["payload"].get("env", {}).get("MOZ_FETCHES", {})
+                        task_def["payload"].get("env", {}).get("MOZ_FETCHES", "{}")
```
hneiva, line 45: super nitty: could also do a check `if env := task_def["payload"].get("env"):....` for arguably more performance. But give this only runs once, it should be fine too.

hneiva (approved):
LGTM

ahal (approved):

PR #223, opened 2024-12-13
refactor: move taskcluster root url definitions to their own file

bhearsum:
This allows them to be imported without depending on any of the imports that `util/integration.py` does. (This is going to become a problem in https://github.com/mozilla-releng/fxci-config/pull/210, when I'm adding a `taskgraph` import there.)


hneiva (approved):
Awesome!

ahal (approved):

PR #225, opened 2024-12-13
refactor: get rid of hardcoded attribute filtering in find_tasks

bhearsum:


taskcluster/test/test_transforms_integration_test.py
```diff
@@ -13,7 +13,7 @@
 from fxci_config_taskgraph.util.integration import (
     FIREFOXCI_ROOT_URL,
     STAGING_ROOT_URL,
-    find_tasks,
+    _find_tasks,
```
hneiva, line 16: Given this is imported in multiple places, I wonder if it's worth either removing the _ prefix (and calling the function something else), or adding a `fxci_config_taskgraph.util.integration.clear_cache` function for these use cases.

hneiva (approved):
LGTM - with a minor/optional improvement recommendation

taskcluster/test/test_transforms_integration_test.py
bhearsum: I'm not sure I see any point in wrapping `cache_clear`. I may be able to find a better name for the private inner function though.

taskcluster/fxci_config_taskgraph/transforms/firefoxci_artifact.py
```diff
@@ -25,8 +25,12 @@ def make_firefoxci_artifact_tasks(config, tasks):
     for task in tasks:
         # Format: { <task id>: [<artifact name>]}
         tasks_to_create = defaultdict(list)
+        include_attrs = task.pop("include-attrs", {})
+        exclude_attrs = task.pop("exclude-attrs", {})
```
ahal, line 29: This can be done in a follow-up, but it might be nice to group `include-attrs`/`exclude-attrs` with individual index paths. E.g:
```
from-decision-indices:
    - path: <index path>
      include-attrs: [<attrs>]
      exclude-attrs: [<attrs>]
```

This would allow a bit more flexibility.

taskcluster/fxci_config_taskgraph/util/integration.py
```diff
@@ -38,14 +41,32 @@ def find_tasks(decision_index_path: str) -> list[dict[str, Any]]:
...
-        if attributes.get("unittest_variant") != "os-integration":
-            continue
+        skip = False
+        for attr, values in dict(include_attrs).items():
+            if not any([attributes.get(attr, "").startswith(v) for v in values]):
+                skip = True
+
+        for attr, values in dict(exclude_attrs).items():
+            if any([attributes.get(attr, "").startswith(v) for v in values]):
+                skip = True
```
ahal, line 67: This feels like it should be using `attrmatch`:
https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/util/attributes.py#L9

But I guess that wouldn't support the exclude case? It might be nice for `attrmatch` to support excluding attributes as well (maybe by prepending `!` to the name or something).

But this is also follow-up territory.

ahal: Actually you could call `attrmatch` twice. Once for `include-attrs`, and a second time for `exclude-attrs` but then negate the return value.

ahal (changes requested):
Requesting changes to use `attrmatch`

taskcluster/fxci_config_taskgraph/util/integration.py
bhearsum: TIL about `attrmatch`. Indeed, let's use that instead.

bhearsum: Switching to this has highlighted that it also may be nice to support prefixes natively in `attrmatch`; for now I've wrapped that portion in a callable that `attrmatch` can use.

ahal: Ah, I missed the `startswith`, good call

ahal (approved):

PR #227, opened 2024-12-17
ci: install ci-admin in Decision tasks

ahal:
This fixes an issue when running the integration-test optimization.


bhearsum (approved):
Thanks!

ahal:
Oops, this should have been draft, think I'm getting close though.

ahal:
`uv` to the rescue!

ahal:
Note that the integration tests are being scheduled on this PR because the optimization keeps them if anything under `/taskcluster` is modified. We can tweak this in a follow-up if we want to get more granular.

PR #237, opened 2024-12-19
Update pinned dependencies

jcristau:


bhearsum (approved):
I see a lot of spam here, but it looks like that's because some of the files in `requirements` somehow lost their `-r` refereneces in the past.

jcristau:
Yeah, looks like they were generated by uv pip compile last time around, while I used pin.sh.

PR #241, opened 2024-12-20
lower queue inactivity timeout for expensive translations workers

bhearsum:
These currently use the default 2 hours, which is very wasteful. 5 minutes ought to be enough for them to reboot and claim another task, if it's available immediately.


hneiva (approved):
LGTM 💸

PR #244, opened 2024-12-27
fix: increase max run time for translations worker pools

bhearsum:
This is to work around https://github.com/taskcluster/taskcluster/issues/7423 in a way that doesn't affect actual max run times or cause confusion in the translations repo.


aerickson (approved):
lgtm

PR #250, opened 2025-01-09
fix: increase queueInActivityTimeout in translations worker pools

bhearsum:
5 minutes causes worker manager to kill the worker. Yarik suggests that the minimum value here should be at least the 20min reclaim period + a 5-10min buffer.


JohanLorenzo (approved):
Rationale and patch make sense to me.

Eijebong (approved):
Makes sense to me, workers don't reclaim their tasks too often.

PR #251, opened 2025-01-09
Create nss pools for win2022 image

jwmoss:


jcristau (commented):
Looks like this is missing a `nss-.*` or `default` case in the maxCapacity setting.

PR #254, opened 2025-01-09
RELOPS-1188: add 22.04 image with gnome-keyring fix, use in test pool

aerickson:
Use a new Ubuntu 22.04 image with a fix for gnome-keyring locking issues.

Based on https://github.com/mozilla-platform-ops/monopacker/pull/153.

see https://mozilla-hub.atlassian.net/browse/RELOPS-1188


aerickson:
Yeah, 1071 had to be removed to get tests passing (it was only used in my test pool).

PR #259, opened 2025-01-14
chore: add staging-reference-browser to projects.yml

bhearsum:
So we can actually run tasks and test things in this repo.

### DIFF
--- a/projects.yml
+++ b/projects.yml
@@ -685,6 +685,37 @@ reference-browser:
       - bump-android-comp
       - nightly
 
+staging-reference-browser:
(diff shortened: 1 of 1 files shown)


projects.yml
```diff
@@ -653,6 +653,38 @@ simple-github:
     taskgraph-actions: true
     trust-domain-scopes: true
 
+# mobile Github repositories
+staging-reference-browser:
```
ahal, line 657: nit: usually the `staging-` repo goes under the main one, and you've accidentally duplicated the comment.

ahal (approved):

PR #260, opened 2025-01-14
Bug 1941649 - Grant generic action scopes for `mozilla` trust domain …

ahal:
…scm_level_*

### DIFF
--- a/grants.yml
+++ b/grants.yml
@@ -2287,6 +2287,7 @@
 - grant:
     - hooks:trigger-hook:project-gecko/in-tree-action-1-generic/*
     - hooks:trigger-hook:project-comm/in-tree-action-1-generic/*
+    - hooks:trigger-hook:project-mozilla/in-tree-action-1-generic/*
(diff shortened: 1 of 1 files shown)


ahal:
I'm going to leave this in draft until others on the team have had a chance to object.

bhearsum (approved):
I had a look at Gecko actions that are using the generic permission. The only one that seemed like it ought not to be usable by anyone with L3 is https://searchfox.org/mozilla-central/source/taskcluster/gecko_taskgraph/actions/openh264.py, which builds and signs the openh264 plugin. It's not like that actually ships anything though, so I'm not terribly concerned about it.

`rebuild_cached_builds` might arguably need to be more protected as well, but that's more for footgun reasons than anything else, and I can still see a fairly strong argument for keeping it generic to keep a wide array of people that can help unstick cached task issues.

ahal:
Just to clarify, we already grant the generic Gecko permissions to everyone with L3 (I think you knew that). I agree that those two should probably have their own scope though.

PR #274, opened 2025-01-31
chore: remove `containerEngine` config

matt-boris:
Generic Worker's config `d2gConfig.containerEngine` will be removed soon with https://github.com/taskcluster/taskcluster/pull/7482.

This change is safe to land now, as `docker` is the default, anyway.

### DIFF
--- a/worker-pools.yml
+++ b/worker-pools.yml
@@ -400,7 +400,6 @@ pools:
             enableInteractive: true
             d2gConfig:
               enableD2G: true
-              containerEngine: docker
             headlessTasks: true
             ed25519SigningKeyLocation: '/etc/generic-worker/ed25519_key'
       minCapacity: 0
@@ -433,7 +432,6 @@ pools:
             enableInteractive: true
             d2gConfig:
               enableD2G: true
-              containerEngine: docker
             headlessTasks: true
             ed25519SigningKeyLocation: '/etc/generic-worker/ed25519_key'
       minCapacity: 0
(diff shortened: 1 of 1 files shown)


bhearsum:
Has `docker` always been the default? (If not, we may need new images before we can land this.)

matt-boris:
> Has `docker` always been the default? (If not, we may need new images before we can land this.)

Good question @bhearsum, I should've given additional context. Since the `containerEngine` field was initially added, it was `docker` by default https://github.com/taskcluster/taskcluster/pull/7265.

bhearsum (approved):
I agree this should be safe in that case! Let's hold off landing until next week on the off chance something wild happens. In the meantime, I'll run integration tests for an additional sanity check.

matt-boris:
Generic Worker v80.0.0 contains this change.

https://github.com/taskcluster/taskcluster/releases/tag/v80.0.0

bhearsum:
I will note that we merged this with failing integration tests. In this case, I'm fairly certain it was unrelated to the PR, but in the future we should either have all green integration tests, or a comment about why it's OK to ignore the failures before merging.

ahal:
Ah sorry, I did look at the failures and it seemed unrelated, so figured it was just due to some edge case in the tests. Should have commented.

matt-boris:
@bhearsum the integration test suite that you started over this past weekend all passed successfully. Then jmoss started again and it looked like instances were getting preempted a bunch.

bhearsum:
Thanks folks, good to know! (I couldn't find the status of mine after they were retriggered....)

PR #279, opened 2025-02-06
[RELOPS-773] Provision new Mac Signers

rcurranmoz:
I’m familiar with creating worker pools like this:

```
project/releng/generic-worker/datacenter-gecko-t-osx-1400-r8:
  description: Datacenter MacOS dedicated developer and automation worker.
  scopes:
    - assume:worker-type:releng-hardware/gecko-t-osx*
    - queue:worker-id:mdc1/mac*
```
However, setting up the new Mac hardware signing machines is a bit more of a mystery to me.

For reference, the current M2 Mac signing hardware designated for this purpose is named something like this (though this can change)

`macmini-m2-44.test.releng.mdc1.mozilla.com`

The goal—perhaps obvious—is to spin up a new Firefox production signing pool on ARM/macOS 14.

Would love any feedback or insights! 😊

Existing signer [pools](https://firefox-ci-tc.services.mozilla.com/provisioners/scriptworker-prov-v1/worker-types/) for reference

### DIFF
--- a/clients.yml
+++ b/clients.yml
@@ -896,6 +896,14 @@ project/releng/scriptworker/v2/mac-signing/prod/firefoxci-mozillavpn-t:
     # TODO: Remove once mac-signing19 is not being used for VPN anymore
     - assume:worker-type:scriptworker-prov-v1/mozillavpn-t-signing-mac
     - queue:worker-id:mozillavpn-t-signing-mac/vpn-mac-v3-signing*
+project/releng/scriptworker/v2/mac-signing/prod/firefoxci-ff-mac14-prod:
+  description: firefox prod signing on macOS14
+  scopes:
+    - assume:worker-type:scriptworker-prov-v1/vpn-depsigning-mac-v1
+    - queue:worker-id:vpn-depsigning-mac-v1/dep-mac-v3-signing*
+    # TODO: Remove once mac-signing19 is not being used for VPN anymore
+    - assume:worker-type:scriptworker-prov-v1/mozillavpn-t-signing-mac
+    - queue:worker-id:mozillavpn-t-signing-mac/vpn-mac-v3-signing*
 project/releng/scriptworker/v2/mac-signing/prod/firefoxci-mozillavpn-3:
   description: mozillavpn mac signing
   scopes:
--- a/clients.yml
+++ b/clients.yml
@@ -873,6 +873,11 @@ project/releng/scriptworker/v2/iscript/tb-prod:
   scopes:
     - assume:worker-type:scriptworker-prov-v1/tb-signing-mac-v1
     - queue:worker-id:signing-mac-v1/tb-mac-v3-signing*
+project/releng/scriptworker/v2/mac-signing/prod/firefoxci-gecko-signing-mac14m2-dep:
+  description: Dep signing worker pool for Firefox on macOS 14 (M2). Handles macOS-specific signing tasks.
+  scopes:
+    - assume:worker-type:scriptworker-prov-v1/gecko-signing-mac14m2-prod
+    - queue:worker-id:gecko-signing-mac14m2-prod/signing-mac14m2-prod*
 project/releng/scriptworker/v2/mac-signing/prod/firefoxci-gecko-signing-mac14m2:
   description: Production signing worker pool for Firefox on macOS 14 (M2). Handles macOS-specific signing tasks.
   scopes:
--- a/clients.yml
+++ b/clients.yml
@@ -873,16 +873,16 @@ project/releng/scriptworker/v2/iscript/tb-prod:
   scopes:
     - assume:worker-type:scriptworker-prov-v1/tb-signing-mac-v1
     - queue:worker-id:signing-mac-v1/tb-mac-v3-signing*
-project/releng/scriptworker/v2/mac-signing/prod/firefoxci-gecko-signing-mac14m2-dep:
+project/releng/scriptworker/v2/mac-signing/dev/firefoxci-gecko-t:
   description: Dep signing worker pool for Firefox on macOS 14 (M2). Handles macOS-specific signing tasks.
   scopes:
-    - assume:worker-type:scriptworker-prov-v1/gecko-signing-mac14m2-prod
-    - queue:worker-id:gecko-signing-mac14m2-prod/signing-mac14m2-prod*
-project/releng/scriptworker/v2/mac-signing/prod/firefoxci-gecko-signing-mac14m2:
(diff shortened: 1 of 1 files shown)


clients.yml
```diff
@@ -896,6 +896,11 @@ project/releng/scriptworker/v2/mac-signing/prod/firefoxci-mozillavpn-t:
     # TODO: Remove once mac-signing19 is not being used for VPN anymore
     - assume:worker-type:scriptworker-prov-v1/mozillavpn-t-signing-mac
     - queue:worker-id:mozillavpn-t-signing-mac/vpn-mac-v3-signing*
+project/releng/scriptworker/v2/mac-signing/prod/firefoxci-ff-mac14m2-prod:
```
ahal, line 899: Sorry, going to be a bit nitpicky:

1. Could you replace `ff` with the trust domain, `gecko`, instead? Please also change in the description.
2. Could you move this above the `mozillavpn` section (around line 863) so it's next to the other Gekco signing scopes.
3. This has `prod` in the name (twice), but we're granting dep scopes. So either the name should be update to dep or we're granting the wrong scopes. I think the second `-prod` at the end isn't necessary.

ahal (changes requested):
Do you know why some of the mac signer clients have `iscript` in the name and some have `mac-signing` in the name? Is `iscript` for the old signers and `mac-signing` for the new ones? If you're unsure, no worries. I like `mac-signing` better myself anyway.

clients.yml
ahal: Seems reasonable to me, but I don't have a strong opinion

rcurranmoz:
> Do you know why some of the mac signer clients have `iscript` in the name and some have `mac-signing` in the name? Is `iscript` for the old signers and `mac-signing` for the new ones? If you're unsure, no worries. I like `mac-signing` better myself anyway.

not 100% on that

@hneiva am I mistaken or do i remember you saying the hosts with `mac-signing` in the name were the actual physical hosts and the iscript 'hosts' were actually instances (in some cases multiple) of iscript running on a machine?

clients.yml
rcurranmoz: I’d prefer to move forward with the production pool for now, as the current Puppet code is designed specifically for it. Once we’ve validated that the configuration works as expected, adapting the code for other pools should be a relatively easy task.

rcurranmoz: @ahal can you look at the latest commit when you get a chance? Thank you

clients.yml
```diff
@@ -873,6 +873,11 @@ project/releng/scriptworker/v2/iscript/tb-prod:
   scopes:
     - assume:worker-type:scriptworker-prov-v1/tb-signing-mac-v1
     - queue:worker-id:signing-mac-v1/tb-mac-v3-signing*
+project/releng/scriptworker/v2/mac-signing/prod/firefoxci-gecko-signing-mac14m2-prod:
```
ahal, line 876: The extra `-prod` at the end still seems unnecessary.

ahal (approved):
Thanks, that looks better!

clients.yml
rcurranmoz: New commit :)

hneiva:
> > Do you know why some of the mac signer clients have `iscript` in the name and some have `mac-signing` in the name? Is `iscript` for the old signers and `mac-signing` for the new ones? If you're unsure, no worries. I like `mac-signing` better myself anyway.
> 
> Not 100% on that
> 
> @hneiva do I remember correctly that hosts with `mac-signing` in their name refer to the actual physical machines, while the ‘hosts’ in iscript are actually instances of iscript running on a machine (sometimes multiple instances per machine)

I'm not 100% sure on the naming, but sounds right. There are hosts that run more than one worker, yes.

rcurranmoz:
@hneiva @ahal new push for dep pool. lmk

clients.yml
```diff
@@ -873,6 +873,16 @@ project/releng/scriptworker/v2/iscript/tb-prod:
   scopes:
     - assume:worker-type:scriptworker-prov-v1/tb-signing-mac-v1
     - queue:worker-id:signing-mac-v1/tb-mac-v3-signing*
+project/releng/scriptworker/v2/mac-signing/prod/firefoxci-gecko-signing-mac14m2-dep:
+  description: Dep signing worker pool for Firefox on macOS 14 (M2). Handles macOS-specific signing tasks.
+  scopes:
+    - assume:worker-type:scriptworker-prov-v1/gecko-signing-mac14m2-prod
+    - queue:worker-id:gecko-signing-mac14m2-prod/signing-mac14m2-prod*
```
ahal, line 880: This looks like a `dev` or `dep` (the naming sucks) client being granted `prod` scopes, so not good.

I'm trying to figure out what the format *should* be based on other clients, and it's a bit of a mess. I think I see mistakes in some of those other clients as well. But as best I can tell, it should be something like:
```
project/releng/scriptworker/v2/mac-signing/dev/firefoxci-gecko-t:
  description: Dep signing worker pool for Firefox on macOS 14 (M2). Handles macOS-specific signing tasks.
  scopes:
    - assume:worker-type:scriptworker-prov-v1/gecko-depsigning-mac14m2
    - queue:worker-id:gecko-depsigning-mac14m2/depsigning-mac14m2*
```

I'm not really sure what the values in puppet are or how much leeway we have to change things here. This is just a first stab.

clients.yml
```diff
@@ -873,6 +873,16 @@ project/releng/scriptworker/v2/iscript/tb-prod:
...
+project/releng/scriptworker/v2/mac-signing/prod/firefoxci-gecko-signing-mac14m2:
+  description: Production signing worker pool for Firefox on macOS 14 (M2). Handles macOS-specific signing tasks.
+  scopes:
+    - assume:worker-type:scriptworker-prov-v1/gecko-signing-mac14m2-prod
+    - queue:worker-id:gecko-signing-mac14m2-prod/signing-mac14m2-prod*
```
ahal, line 885: Similarly, I'd make this one something like:
```
project/releng/scriptworker/v2/mac-signing/prod/firefoxci-gecko-3:
  description: Production signing worker pool for Firefox on macOS 14 (M2). Handles macOS-specific signing tasks.
  scopes:
    - assume:worker-type:scriptworker-prov-v1/gecko-signing-mac14m2
    - queue:worker-id:gecko-signing-mac14m2/signing-mac14m2*
```

rcurranmoz:
changes made @ahal thanks for your help

rcurranmoz:
@ahal @hneiva how does this look?

hneiva (approved):
LGTM

ahal (approved):

PR #287, opened 2025-02-11
feat: add lando scriptworker client for dev

bhearsum:
Note that this is getting added in `clients-interpreted.yml` for Gecko, which is something we haven't done up until now. I have mixed feelings about this as these clients are shared across multiple projects, but I _think_ having consistency in the client entries is worthwhile.

### DIFF
--- a/clients-interpreted.yml
+++ b/clients-interpreted.yml
@@ -174,3 +174,15 @@
       - project:
             level: 3
             feature: [beetmover-get-artifact-scope]
+
+# Clients for landoscript workers
+- client:
+      project/releng/scriptworker/v2/lando/dev/firefoxci-{trust_domain}-1:
+          description: "{trust_domain} level 1 nonprod lando scriptworker"
+          scopes:
+              - queue:claim-work:scriptworker-k8s/{trust_domain}-1-lando-dev
+              - queue:worker-id:{trust_domain}-1-lando-dev/{trust_domain}-1-lando-dev-*
+  for:
+      - project:
+            level: 1
+            feature: [lando]
--- a/projects.yml
+++ b/projects.yml
@@ -405,6 +405,7 @@ try:
     hg-push: true
     treeherder-reporting: true
     trust-domain-scopes: true
+    lando: true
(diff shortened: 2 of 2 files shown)


clients-interpreted.yml
```diff
@@ -174,3 +174,15 @@
...
             level: 3
             feature: [beetmover-get-artifact-scope]
+
+# Clients for landoscript workers
+- client:
+      project/releng/scriptworker/v2/lando/dev/firefoxci-{trust_domain}-1:
+          description: "{trust_domain} level 1 nonprod lando scriptworker"
+          scopes:
+              - queue:claim-work:scriptworker-k8s/{trust_domain}-1-lando-dev
+              - queue:worker-id:{trust_domain}-1-lando-dev/{trust_domain}-1-lando-dev-*
```
ahal, line 184: So sharing a single client across projects is harmless as long as the scopes don't have `{trust_project}` anywhere in them. As it stands, even if we did split clients out by project they would all have the same scopes. So there would be no security benefit of doing that.

However, maybe there should be per-project scopes? I guess it ultimately comes down to Lando. Will they be providing us with a single access token that can land to every branch in the new repo? Or will there be per-branch tokens?

If it's the former, then a single scope is fine. If it's the latter, we might want to separate it out here as well.

ahal (approved):

clients-interpreted.yml
bhearsum: > However, maybe there should be per-project scopes? I guess it ultimately comes down to Lando. Will they be providing us with a single access token that can land to every branch in the new repo? Or will there be per-branch tokens?

The token or tokens we'll be getting are unclear at this time. It's also unclear if we'll have extra scopes that may differ by project.

However...we will need separate workloads per trust domain as I understand things, unless you're suggesting that we have a single `landoscript` workload that doesn't include trust domain anywhere? (I'm not even sure how that would work for chain of trust verification though...)

Obviously we could share the same client id across multiple workloads; I'm not sure if that's a good, bad, or neutral thing...

PR #292, opened 2025-02-21
chore: pick up new ubuntu 24.04 headless image

bhearsum:
The main change here is generic-worker 81.0.3, which includes interactive task improvements and chain of trust d2g fixes compared to the current image.

### DIFF
--- a/worker-images.yml
+++ b/worker-images.yml
@@ -98,7 +98,7 @@ ubuntu-2404-headless-alpha:
   fxci-level1-gcp: projects/taskcluster-imaging/global/images/gw-fxci-gcp-l1-2404-amd64-headless-googlecompute-alpha
 ubuntu-2404-headless:
   ## Headless Image for Ubuntu 24.04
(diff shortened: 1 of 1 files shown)


bhearsum:
test failure seems to be a bug in the translations integration tests; I'm looking into it.

bhearsum:
integration tests should be fixed with https://github.com/mozilla-releng/fxci-config/pull/294; I'll rebase when that lands

bhearsum:
The errors here are most certainly unrelated to the new image - it looks like a legitimate issue in the translations repo around python dependencies. The original ones were on [tasks like this](https://stage.taskcluster.nonprod.cloudops.mozgcp.net/tasks/RcyQqDyfRMaQ_EmgDuk-dQ), which use the monopacker images still. I tried to force things along, but that resulted in downstream failures due to missing artifacts.

IMO we've gotten deep enough into the pipeline to verify the new images - almost all of the tasks that use them. I'm going to land this; we can revert on the off chance something comes up.

PR #294, opened 2025-02-21
fix: ignore new and old names for docker image tasks in translations integration tests

bhearsum:
These tasks changed names when Translations picked up a new Taskgraph version. At the moment we're in a state where both are present in the most recent cron run (some docker images have been rebuilt, others haven't); we need to keep both in here for awhile to avoid accidentally trying to rebuild them in staging (which doesn't work, and is wasteful).

This should fix the issue with integration tests I noticed in #292.

### DIFF
--- a/taskcluster/kinds/firefoxci-artifact/kind.yml
+++ b/taskcluster/kinds/firefoxci-artifact/kind.yml
@@ -59,6 +59,6 @@ tasks:
       # and also because they live forever, and depend on docker images which
       # expire, which causes us to try to mirror expired docker image tasks
       # into the staging cluster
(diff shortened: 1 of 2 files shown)


bhearsum:
The first run of integration tests failed due to https://bugzilla.mozilla.org/show_bug.cgi?id=1949852. I (think) I've hacked around it in the short term by adding some extra scopes to the staging cluster. We'll see how a rerun of the decision task goes.

bhearsum:
hack worked....this should be good to merge. the thing i was hacking around is fixed in https://github.com/mozilla-releng/fxci-config/pull/295

ahal (approved):

PR #306, opened 2025-03-05
Add gecko-t/t-linux-2404-headless-ssd-alpha pool

jcristau:
We'd like to start setting up local ssds in worker images with g-w/d2g, so add a pool with a scratch disk.


hneiva (approved):
LGTM

PR #311, opened 2025-03-12
linux-2404-headless-ssd-alpha: move generic-worker caches/downloads t…

jcristau:
…o /home

Per https://github.com/taskcluster/taskcluster/issues/7580 this needs to be under the same mount point as the tasks dir.  As an added benefit this moves them to the local ssd.

### DIFF
--- a/worker-pools.yml
+++ b/worker-pools.yml
@@ -1623,6 +1623,8 @@ pools:
           config:
             enableInteractive: true
             headlessTasks: true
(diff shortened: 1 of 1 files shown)


jcristau:
As far as I can tell the caches dir and its parents should get created as needed (https://github.com/taskcluster/taskcluster/blob/536651f75f14c1cc6d2dd494d530ab0bf91ffc4b/workers/generic-worker/mounts.go#L162), so hopefully this is sufficient.

PR #313, opened 2025-03-14
[RELOPS-773] Update m2 Mac signers scopes + missing grants + comm vs tb 

hneiva:
Switching it to prefix `dep-` so it matches old configs and renames the tb workers to use the trust domain (comm) instead of just tb

### DIFF
--- a/grants.yml
+++ b/grants.yml
@@ -312,6 +312,7 @@
     - queue:create-task:low:lambda/gecko-t-*
     - queue:create-task:low:scriptworker-prov-v1/depsigning-mac-v1
     - queue:create-task:low:scriptworker-prov-v1/depsigning-mac-v1-dev
+    - queue:create-task:low:scriptworker-prov-v1/dep-{trust_domain}-signing-mac14m2
     - queue:get-artifact:releng/partner/*
     # TODO(bug 1824855): move mobile secrets into the gecko namespace
     - secrets:get:project/mobile/firefox-android/android-components/firebase
@@ -368,6 +369,7 @@
     - queue:create-task:low:bitbar/gecko-t-*
     - queue:create-task:low:scriptworker-prov-v1/depsigning-mac-v1
     - queue:create-task:low:scriptworker-prov-v1/depsigning-mac-v1-dev
+    - queue:create-task:low:scriptworker-prov-v1/dep-{trust_domain}-signing-mac14m2
     - queue:get-artifact:releng/partner/*
   to:
     - projects:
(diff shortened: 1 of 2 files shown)


grants.yml
```diff
@@ -312,6 +312,8 @@
     - queue:create-task:low:lambda/gecko-t-*
     - queue:create-task:low:scriptworker-prov-v1/depsigning-mac-v1
     - queue:create-task:low:scriptworker-prov-v1/depsigning-mac-v1-dev
+    - queue:create-task:low:scriptworker-prov-v1/dep-gecko-signing-mac14m2
+    - queue:create-task:low:scriptworker-prov-v1/gecko-signing-mac14m2
```
jcristau, line 316: This looks like it's giving prod signing access to l1 repos?

grants.yml
```diff
@@ -393,7 +395,9 @@
     - project:comm:thunderbird:releng:treescript:action:push
     - project:comm:thunderbird:releng:treescript:action:tagging
     - project:comm:thunderbird:releng:treescript:action:version_bump
-    - queue:create-task:low:scriptworker-prov-v1/tb-depsigning-mac-v1
+    - queue:create-task:low:scriptworker-prov-v1/tb-signing-mac-v1
```
jcristau, line 398: Likewise.

grants.yml
```diff
@@ -2013,6 +2017,8 @@
     - queue:create-task:highest:adhoc-t/*
     - queue:create-task:highest:scriptworker-prov-v1/{trust_domain}-signing-mac-dev
     - queue:create-task:highest:scriptworker-prov-v1/{trust_domain}-{level}-*
+    - queue:create-task:highest:scriptworker-prov-v1/{trust_domain}-signing-mac14m2*
```
jcristau, line 2020: Likewise.

jcristau (changes requested):

grants.yml
hneiva: Nice catch! Thanks!

grants.yml
ahal: I don't think this is fixed? This is a prod signing scope, granted to an L1 context.

ahal (changes requested):

grants.yml
hneiva: I got a bit confused since the adhoc worker was both dev and prod, but this makes sense. Moved the prod scope to the master+action grant.

grants.yml
```diff
@@ -1776,6 +1780,7 @@
     - queue:create-task:highest:scriptworker-k8s/{trust_domain}-{level}-*
     - queue:create-task:highest:scriptworker-k8s/{trust_domain}-t-*
     - queue:create-task:highest:scriptworker-prov-v1/{trust_domain}-t-*
+    - queue:create-task:highest:scriptworker-prov-v1/{trust_domain}-signing-mac14m2
```
ahal, line 1783: You'll also need to remove this one. Without the `job:` key below, it'll default to everywhere which includes pull requests :)

ahal (changes requested):

grants.yml
hneiva: Mmmmm right. I've moved it down to main/release only

ahal (approved):
Thanks, lgtm!

PR #315, opened 2025-03-18
[RELOPS-1367] Initial support for ubuntu 24.04 arm64 headless image

jwmoss:


worker-pools.yml
```diff
@@ -1666,6 +1666,30 @@ pools:
...
+      - pool-group: gecko-t
+    config:
+      worker-config:
+        genericWorker:
+          config:
+            enableInteractive: true
+            headlessTasks: true
+      minCapacity: 0
+      maxCapacity: 100
+      implementation: generic-worker/linux-d2g
```
bhearsum, line 1684: I assume you intentionally enabled d2g here? (I'm not sure what this pool will be used for, so I'm not sure whether it's correct.)

gabrielBusta: @bhearsum I am going to run some test harnesses for the ARM64 builds on these. We are piggy backing on the x64 image that will run d2g

bhearsum: I'll defer to you on whether or not this is correct, in that case!

jcristau (approved):

gabrielBusta (approved):

markcor (approved):

PR #317, opened 2025-03-19
add in-tree action hooks for the mozilla trust domain (bug 1955012)

jcristau:


jcristau:
diff: https://gist.github.com/jcristau/ddad41d4069ea8fb62d6cfc7dbab86d1

There's some noise from launchconfigs getting reordered, but other than that it looks sane, I think?

bhearsum (approved):
I agree, this looks sane. (I'm going to see if I can go fix the non-determinism in the launch configs....)

PR #318, opened 2025-03-19
fix: make worker pool generation deterministic

bhearsum:
In https://github.com/mozilla-releng/fxci-config/pull/317 we noticed some seemingly unrelated changes to worker pools in a taskgraph diff. This is almost certainly because we're not sorting regions/AZs/instance types when generating these.

It's difficult to fully verify this, but I did manage to get three identical diffs locally with this patch, and the minimal diff (ie: no additions/removals from reordering) when making a minor change to a worker pool.


gabrielBusta (approved):
Nice, I've noticed stuff like this too. Never looked into it.

jcristau (approved):
Thanks!

PR #319, opened 2025-03-21
Remove unused gecko-3/t-linux* pools (bug 1955562)

jcristau:
We moved the android x86 pgo runs to b-linux-kvm-gcp, so nothing is running there anymore.


ahal (approved):
Thanks!

PR #320, opened 2025-03-21
Add t-linux-docker{,-noscratch} worker pools (bug 1955009)

jcristau:


worker-pools.yml
```diff
@@ -2986,6 +3020,37 @@ pools:
                 sourceImage: <image>
               type: PERSISTENT
           machine_type: c2-standard-4
+  - pool_id: '{pool-group}/t-linux-2404-d2g-noscratch'
```
ahal, line 3023: I implemented a "suffix" mechanism in some other pools to handle cases like this "noscratch" variant. It might be worth re-using that to cut down on config duplication.. however in this case I think it likely would need some code changes to go along with it. So feel free to leave it as a follow-up.

worker-pools.yml
```diff
@@ -2958,6 +2958,40 @@ pools:
               type: SCRATCH
               interface: NVME
           machine_type: c2-standard-4
+  - pool_id: '{pool-group}/t-linux-2404-d2g'
```
ahal, line 2961: While we're standardizing on new pool names.. should we change this from `linux-2404` -> `ubuntu-2404`? It's always mildly bugged me.

ahal, line 2961: Another thought...

Most consumers are not going to know what `d2g` is. Should we make the suffix `-docker` instead? I think this might be slightly more confusing for us, but much less confusing for everyone else.

I'm on the fence with this one, but think I'm leaning towards `-docker`.

worker-pools.yml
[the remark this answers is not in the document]
jcristau: Hmm, I don't really want to be mixing up tests and builds on the same pools, so I kind of like having that purpose prefix.

jcristau: Thanks, will keep that in mind.

worker-pools.yml
jcristau: Maybe.  I don't really like having the 2404 in there to be honest, in theory it's not relevant to the consumers since everything runs in docker.  But I was struggling to come up with a name for `t-linux-{,x}large-gcp`'s successor...

jcristau: Makes sense, thanks.

worker-pools.yml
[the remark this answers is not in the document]
ahal: They wouldn't using the same pool because of the different pool-groups. If we want to carve out separate pools for caching purposes, we can use `{suffix}` for that. Or if you really want it at the beginning, I suppose we could do `{prefix}`, but keeping all these strings at the end feels cleaner.

Consider, both `t-` and `-source` are only there to keep groups of related tasks running together. Why is one at the beginning and the other at the end?

worker-pools.yml
ahal: Makes sense, what's stopping us from dropping the `2404` exactly? Doesn't the `d2g` already differentiate it enough?

worker-pools.yml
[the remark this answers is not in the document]
ahal: Just for an example taking all the other comment threads into account, the final resolved version of the pool name could be something like:
```
gecko-1/linux-docker-noscratch-test
```

I'd also advocate getting rid of `gecko-t` and just use `gecko-1`

(I have lots of opinions on these pool names, you opened Pandoras box :laughing: )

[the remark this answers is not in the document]
ahal: I forgot `gecko-t` works across both levels, so ignore that

worker-pools.yml
gabrielBusta: Yeah that would make more sense. Can pools that enable d2g still run tasks in the worker without using docker? Ah, I guess we only care to enable this on pools we are going to be running docker tasks on.

worker-pools.yml
```diff
@@ -2958,6 +2958,40 @@ pools:
               type: SCRATCH
               interface: NVME
           machine_type: c2-standard-4
+  - pool_id: '{pool-group}/t-linux-docker'
```
bhearsum, line 2961: I like the naming! I assume `-docker` implies that `disableNativePayloads` should be `true`?

bhearsum (approved):

worker-pools.yml
jcristau: Yup.

jcristau: (It is, as far as I can tell)

bhearsum: This makes me wonder if we ought to have an implementation for these types of pools that would help enforce this (it wouldn't be a guarantee, but it would make it more difficult to forget to set it...).

jcristau:
> Should we use Ubuntu vs Linux in the name?

My preference would be to keep it as linux, since consumers should largely not have to care about the ubuntu part, just that they get to run things inside docker.

PR #324, opened 2025-03-25
chore: remove unused translations non-d2g cpu workers

bhearsum:
We switched these workers to new worker pools in https://github.com/mozilla/translations/pull/949. Unless there's some older branches still used for training that haven't picked up this patch, these are now unused.

@eu9ene or @gregtatum - are either of you aware of any branches that were made prior to mid-December that might still need these workers? (If so, we can hold off cleaning this up for awhile longer.)


ahal (approved):

gregtatum:
I have none.

bhearsum:
I'm going to go ahead and merge this. We can revert it if the workers are still needed for some reason.

PR #326, opened 2025-03-26
Bug 1956630 - Adjust `t-linux-vm-2204-wayland` idleTimeoutSecs to som…

ahal:
…ething reasonable

### DIFF
--- a/worker-pools.yml
+++ b/worker-pools.yml
@@ -1439,7 +1439,7 @@ pools:
             livelogExecutable: /home/ubuntu/.taskcluster/livelog
             taskclusterProxyExecutable: /home/ubuntu/.taskcluster/taskcluster-proxy
             numberOfTasksToRun: 1  # gw simple uses 1 to exit after one task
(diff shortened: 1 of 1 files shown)


ahal:
I'm tempted to also cut down `maxWorkers` a bit. Looks like we get between 2-4k tasks in these pools per work day. 500 feels like it's probably overkill? Given there's so much overhead in these pools, it's probably not a bad thing to allow a bit of a pending queue to grow here.

Anyone have thoughts?

bhearsum:
> I'm tempted to also cut down `maxWorkers` a bit. Looks like we get between 2-4k tasks in these pools per work day. 500 feels like it's probably overkill? Given there's so much overhead in these pools, it's probably not a bad thing to allow a bit of a pending queue to grow here.
> 
> Anyone have thoughts?

Depending on how the load is spread out that could end up better or worse, but I don't see a reason not to give it a try.

bhearsum (approved):

PR #334, opened 2025-04-01
clients: add non-ldap client for jcristau

jcristau:
When logging in to the UI via auth0 I end up with an auth certificate larger than 4k (due to the list of ldap and mozillians groups), causing breakage because it's included in bewit tokens in URLs and the Authorization header.  Add a static client with star scopes as a workaround.


bhearsum (approved):
This was discussed a bunch elsewhere. tl;dr - it's not great to have a non-expiring client for a specific user with such high privileges, but as a short/medium term workaround it's OK. Ideally @jcristau is going to find a way to reduce the size of the certificate.

PR #335, opened 2025-04-01
chore: bump ubuntu headless 24.04 image to pick up new generic-worker version

bhearsum:
This moves us from 81.0.3 to 83.4.0, most notably picking up the new `volume` artifact type (https://github.com/taskcluster/taskcluster/issues/7594) and OOM monitoring (https://github.com/taskcluster/taskcluster/issues/6464).

The former is a blocking issues for moving translations GPU workers over to this image.


bhearsum:
Integration test failure should be fixed by https://github.com/mozilla/translations/pull/1069.

PR #336, opened 2025-04-01
feat: Bug 1957757 Add mobile trigger-hook scope to shipit clients

hneiva:


hneiva:
Should fix the error when triggering a promote task from shipit:

---

Client ID project/releng/shipit/staging does not have sufficient scopes and is missing the following scopes:
```hooks:trigger-hook:project-mobile/in-tree-action-1-release-promotion/1acbb25a07```
This request requires the client to satisfy the following scope expression:
`hooks:trigger-hook:project-mobile/in-tree-action-1-release-promotion/1acbb25a07`
\---
* method:     triggerHook
* errorCode:  InsufficientScopes
* statusCode: 403
* time:       2025-04-01T20:19:38.791Z

bhearsum:
Can you provide some additional context here on why this is needed now? I'm guessing it's related to iOS automation? (Some sort of bug or cross link would be helpful....)

bhearsum (approved):

PR #347, opened 2025-04-17
chore: bump ubuntu headless 24.04 image to pick up new generic-worker version

matt-boris:
https://docs.taskcluster.net/docs/changelog?version=v83.5.6&audience=WORKER-DEPLOYERS


bhearsum:
@jwmoss - Do we know if there's anything besides generic-worker being updated here? (If it's only generic worker, I think it's probably fine to take this despite the integration test failure.)

bhearsum (approved):

PR #354, opened 2025-04-25
Bug 1950585 - part 4: Fix c3d-standard16-lssd as they must have 1 loc…

JohanLorenzo:
…al SSD and they're not present in us-west1

I tried to be smart in https://github.com/mozilla-releng/fxci-config/pull/349 by factorizing the config between `c2d` and `c3d`. However, their configs actually differ too much. 

`c3d` erred out: 

```
The selected machine type(c3d-standard-16-lssd) should have [1] local SSD(s).
```

```
Invalid value for field 'resource.machineType': 'zones/us-west1-b/machineTypes/c3d-standard-16-lssd'. Machine type with name 'c3d-standard-16-lssd' does not exist in zone 'us-west1-b'.
```

(same error with other subregions of `us-west1`)

The Google Console shows where this machine type is available: 

image

### DIFF
--- a/src/ciadmin/check/check_worker_pools.py
+++ b/src/ciadmin/check/check_worker_pools.py
@@ -153,9 +153,8 @@ async def check_gcp_ssds():
     worker_pools = await WorkerPoolConfig.fetch_all()
     ignore = (
         # Bug 1962119: Let's keep the number of local SSDs the same between
-        # c2 and its AMD counterparts.
+        # c2 and its AMD counterpart.
         "gecko-1/b-linux-gcp-bug1962119-c2d",
-        "gecko-1/b-linux-gcp-bug1962119-c3d",
     )
     errors = []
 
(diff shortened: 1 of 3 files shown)


src/ciadmin/check/check_worker_pools.py
```diff
@@ -153,9 +153,8 @@ async def check_gcp_ssds():
     worker_pools = await WorkerPoolConfig.fetch_all()
     ignore = (
         # Bug 1962119: Let's keep the number of local SSDs the same between
-        # c2 and its AMD counterparts.
+        # c2 and its AMD counterpart.
         "gecko-1/b-linux-gcp-bug1962119-c2d",
```
jcristau, line 157: If we can have fewer SSDs on these then we should, IMO.

jcristau (approved):

src/ciadmin/check/check_worker_pools.py
JohanLorenzo: Noted. I'll try to remain as close as possible to the Intel config in order to limit the number of moving parts for now.

PR #359, opened 2025-04-29
fix: update lando repo names to match reality

bhearsum:
I learned yesterday that these are not what I thought they were. The new names match what is in https://phabricator.services.mozilla.com/diffusion/ (`firefox-main` doesn't exist yet, but it will in the near future).


jcristau (approved):
I guess that makes sense.

PR #361, opened 2025-04-30
ci: mirror translations fetch tasks in fxci integration tests

ahal:
Bug: 1961216

I'm not really sure if this is the right fix, as I don't fully understand why fetches were previously excluded. It does seem to get things working for me locally though.

It's worth noting we don't exclude fetches for the gecko tests, nor do we exclude toolchains for translations (which have all the same downsides as fetches per the comment I removed).


ahal:
I guess this started failing because this was the first task in translations that depends on both fetch and non-fetch tasks in its MOZ_FETCHES env.

I confess to be a bit confused about how fetches worked before (without mirroring them over to stage).

taskcluster/kinds/firefoxci-artifact/kind.yml
```diff
@@ -55,10 +55,7 @@ tasks:
       # tasks they normally would be.
       # docker-image tasks are excluded because there's no value in rerunning
       # them as part of integration tests.
-      # fetch tasks are excluded because there's no value in rerunning them,
-      # and also because they live forever, and depend on docker images which
-      # expire, which causes us to try to mirror expired docker image tasks
-      # into the staging cluster
-      - ^(?!(Decision|Action|PR action|build-docker-image|docker-image|fetch)).*
+      - ^(?!(Decision|Action|PR action|build-docker-image|docker-image)).*
```
bhearsum, line 58: Based on the comment and some vague memories I have, I think this a time bomb: [one of the recreated fetch tasks](https://stage.taskcluster.nonprod.cloudops.mozgcp.net/tasks/CTRXpDdtR-q6qqq_ySgQlw) depends on [a mirrored docker-image task](https://stage.taskcluster.nonprod.cloudops.mozgcp.net/tasks/JlWc5PtjT-6MiACiyeyZbw). The [fxci cluster task being mirrored](https://firefox-ci-tc.services.mozilla.com/tasks/MzhBZaOMT0Ofj51AW2FraQ) expires before the fetch tasks. Once we reach this expiry point, integration tests will fail because they can't mirror this image anymore.

(I think this is also a problem for Gecko integration tests, but I assume we've always rebuilt the `fetch` docker-image task often enough for it not to actually happen...)

bhearsum: I should also add: I'm not sure there's a better solution than the one in this patch; but it would force us to artificially bust caches in translations every year, which is not ideal (but not as much of a problem as it used to be...).

jcristau: Once we reach the docker image tasks's expiry, won't it be replaced by a fresh one in translations?

jcristau (approved):

PR #363, opened 2025-05-01
refactor(ciadmin): handle worker-defaults better while generating poo…

ahal:
…l config


jcristau (commented):
Does this change/fix anything in the output?  I couldn't really tell from the patch.

ahal:
Yes, it should be an identical diff, though it will allow keying parts of the worker-defaults by implementation / provider.

When running the diff locally, I see the launchConfigs are in a different order.. but I think this was caused by Yarik's launchConfig changes that recently landed.

bhearsum (approved):

PR #366, opened 2025-05-06
fix(landoscript): grant try scopes for all staging-* lando repositories

bhearsum:
We're switching to the repositories being created in https://github.com/mozilla-conduit/lando/pull/316, which will allow us to have a RelEng controlled repository hooked up to stage Lando.

### DIFF
--- a/grants.yml
+++ b/grants.yml
@@ -3067,8 +3067,9 @@
 # allow all actions to be run through any means.
 - grant:
     - project:releng:lando:action:*
-    # this dev/test repo will probably change; but it's what we have for now
-    - project:releng:lando:repo:ff-test-dev
+    # there are staging versions of all production repos; allow try access
+    # to all of them
+    - project:releng:lando:repo:staging-*
   to:
     - project:
         alias: try


grants.yml
```diff
@@ -3067,8 +3067,9 @@
 # allow all actions to be run through any means.
 - grant:
     - project:releng:lando:action:*
-    # this dev/test repo will probably change; but it's what we have for now
-    - project:releng:lando:repo:ff-test-dev
+    # there are staging versions of all production repos; allow try access
+    # to all of them
+    - project:releng:lando:repo:staging-*
```
ahal, line 3072: Just to clarify, these are repos that are configured in lando, and not any arbitrary repo that starts with `staging-*` across Github, right?

ahal (approved):

grants.yml
bhearsum: That's right. This is from the stage admin panel:
image

There's no technical reason that a non-RelEng repo couldn't be in that list in the future...but I don't think our automation account would have access to it (and if it did...that seems like it would a problem beyond the scope of fxci-config).

PR #367, opened 2025-05-09
Bug 1964918 - add missing scope for index-tasks on "trunk"

jcristau:
When index entries are added by an index-task instead of directly via routes on the task itself (as happens when the task has more than 10 routes), that index-task needs the index:insert-task scope.

### DIFF
--- a/grants.yml
+++ b/grants.yml
@@ -1014,6 +1014,7 @@
 
 - grant:
     - queue:route:index.{trust_domain}.v2.trunk.revision.*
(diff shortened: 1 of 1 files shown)


Eijebong (approved):
Assuming you meant "via routes on the task" and not "scopes on the task". But this sounds reasonable

jcristau:
> Assuming you meant "via routes on the task" and not "scopes on the task". But this sounds reasonable

Quite right.  Updated.

PR #381, opened 2025-05-16
bug 1966916: fix: use github token when fetching .taskcluster.yml files

bhearsum:


jcristau (approved):

src/ciadmin/generate/tcyml.py
```diff
@@ -57,6 +58,10 @@ async def get(repo_path, repo_type="hg", revision=None, default_branch=None):
             retry_options=ExponentialRetry(attempts=5, statuses={404}),
         )
         headers = {"User-Agent": USER_AGENT}
+        # TODO: when we no longer need to support hg.mozilla.org here, we can
+        # switch this to simple-github rather than doing this by hand.
```
ahal, line 62: To be fair we could have a separate code path that uses simple-github now.. but this is fine :)

PR #384, opened 2025-05-20
Bug 1967254 - update generic-worker on 24.04 headless image to 84.0.1

jcristau:


jcristau:
taskcluster integration

PR #386, opened 2025-05-20
Add the required config to run cronjob on the firefox ios repo

Eijebong:
Need this to be able to test automating beta releases

### DIFF
--- a/projects.yml
+++ b/projects.yml
@@ -773,9 +773,13 @@ staging-firefox-ios:
     mobile-testrail: true
     shipit: true
     taskgraph-actions: true
+    taskgraph-cron: true
     treeherder-reporting: true
     treescript-phase: true
     trust-domain-scopes: true
+  cron:
+    targets:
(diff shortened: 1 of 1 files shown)


bhearsum (approved):
Are there any humans (other than RelEng) that will need scopes to trigger this hook? We might want this hook added to https://github.com/mozilla-releng/fxci-config/blob/54989e5351fde663564ac96e82aec23e91a457ca/grants.yml#L732-L736, or maybe a different one if the humans are not RelMan.

(I think for automation purposes you won't need explicit scope grants...but I haven't dug deeply enough to confirm that.)

Eijebong:
This is for the staging repo, I'm not expecting anyone except releng to ever need access to it. Once we transfer the hook to the main repo then we'll probably want relman to be able to trigger it manually too. But for now this is for testing only (and I'm expecting a lot of other things to go wrong 🙃)

PR #387, opened 2025-05-20
ci: fix Gecko os-integration tests

ahal:
The tasks on the Gecko side were refactored to not be a variant, but I forgot that there was a filter based on that over here on the fxci-config side.

I don't think it's important that we filter the "exact" same set of tasks. This is mainly just to avoid picking up tasks that aren't on the "happy" path. Builds and tests are mostly all we care about, so start with that. We can adjust later if more are needed.


taskcluster/kinds/firefoxci-artifact/kind.yml
```diff
@@ -35,8 +35,10 @@ tasks:
     decision-index-paths:
       - gecko.v2.mozilla-central.latest.taskgraph.decision-os-integration
     include-attrs:
-      unittest_variant:
-        - os-integration
+      kind:
```
jcristau, line 38: This doesn't seem to work?  (i.e. the kind isn't an attribute)

jcristau (approved):
Sorry for the lag.

ahal:
I don't think I meant for this to be reviewed yet, it's still in draft. It looks like I may have broken all the translations tasks with this. Maybe things are working without this now?

ahal:
Ok, there's still a failure in translations, but at least Gecko is all working. I think it's ok to land this as is for now and dig into the translation failure in a follow-up. I believe it's unrelated to this PR.

PR #391, opened 2025-05-21
Add scopes to create ios releases on mobile-1-dev shipit scriptworker

Eijebong:
I'm not sure this is correct, I don't understand why we have to have 3 of the same scopes (no prefix/dev/staging) but that seems to be what other shipit dev workers do 🤷

### DIFF
--- a/clients.yml
+++ b/clients.yml
@@ -1049,6 +1049,13 @@ project/releng/scriptworker/v2/shipit/prod/firefoxci-gecko-3:
     - project:releng:services/shipit_api/update_release_status
     - queue:claim-work:scriptworker-k8s/gecko-3-shipit
     - queue:worker-id:gecko-3-shipit/gecko-3-shipit-*
+project/releng/scriptworker/v2/shipit/dev/firefoxci-mobile-1:
+  description: "mobile level 1 nonprod shipit scriptworker"
+  scopes:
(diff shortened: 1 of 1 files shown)


bhearsum (approved):
> I'm not sure this is correct, I don't understand why we have to have 3 of the same scopes (no prefix/dev/staging) but that seems to be what other shipit dev workers do 🤷

This is probably for historic reasons that no longer hold true. As far as I can tell, there is no environment marker in the scope prefix anymore: https://github.com/mozilla-releng/shipit/blob/cb43f6f9b01d26188996456a08d08b5ba8a677af/api/src/shipit_api/common/config.py#L17. In fact, tracing it all the way back [release-services](https://github.com/mozilla/release-services/blame/114f9044c287717869e60925c50470c2af1ef482/src/shipit/api/shipit_api/config.py) I don't see this _ever_ having been used.

The environment marked clients appear to been first added in https://github.com/mozilla-releng/fxci-config/commit/654331ae8bfdcf646fe9c3572541e398ee291e05, which contains no hints at why.

If you're up for it, you could try removing the `dev` and `staging` ones to validate this interpretation (and then we can remove them for other clients if it works...). Totally optional.

Eijebong:
Removed the dev/staging ones, we'll see how that goes, thanks for digging this up

PR #392, opened 2025-05-21
Remove unused prefixed shipit_api scopes

Eijebong:
We found out in #391 that those scopes were most likely not used. After trying out adding firefox-ios without those staging/dev scopes, I can confirm that they are indeed not useful.


Eijebong:
FTR, I tested #391 here https://firefox-ci-tc.services.mozilla.com/tasks/AoQpS5NaS7i02iwJ_Np_rQ/runs/1, first run was before, second run after and the result went from 403 to 400 (because the task is wrong, but it went through auth, at least for add_release, I'm fine with waiting on this until I get the full job working though if you want to be safe)

bhearsum:
> FTR, I tested #391 here https://firefox-ci-tc.services.mozilla.com/tasks/AoQpS5NaS7i02iwJ_Np_rQ/runs/1, first run was before, second run after and the result went from 403 to 400 (because the task is wrong, but it went through auth, at least for add_release, I'm fine with waiting on this until I get the full job working though if you want to be safe)

Yeah, let's wait for that confirmation please!

Eijebong:
It worked 🥳 Created a release and scheduled a ship phase here https://firefox-ci-tc.services.mozilla.com/tasks/MsmSi_fDQ-SwjwF45gF45w/runs/0

bhearsum (approved):
Thank you for cleaning this up!

PR #395, opened 2025-05-22
Bug 1967967 - disable code coverage cron triggers

jcristau:


jcristau:
@marco-c do we want to keep project-relman/code-coverage-cron-testing and project-relman/code-coverage-cron-production firing still?

marco-c:
Those two are not so problematic, but it doesn't hurt disabling them too.

JohanLorenzo (approved):
Okay to deactivate them while the coverage ingestion is broken.

PR #397, opened 2025-05-23
bug 1967745: feat: support public, token, and app clients in tcyml.py for github repositories

bhearsum:
This _should_ fix the staging apply in os-integration tests.

### DIFF
--- a/src/ciadmin/generate/tcyml.py
+++ b/src/ciadmin/generate/tcyml.py
@@ -95,7 +95,9 @@ async def get(repo_path, repo_type="hg", revision=None, default_branch=None):
             params = {"ref": revision}
 
             async with client_cls() as client:
-                response = await client.request("GET", endpoint, headers=headers, params=params)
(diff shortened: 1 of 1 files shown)


src/ciadmin/generate/tcyml.py
```diff
@@ -27,53 +29,85 @@ async def get(repo_path, repo_type="hg", revision=None, default_branch=None):
...
+            elif "GITHUB_APP_ID" in os.environ and "GITHUB_APP_PRIVKEY" in os.environ:
+                client_cls = partial(
+                    AppClient,
+                    os.environ["GITHUB_APP_ID"],
+                    os.environ["GITHUB_APP_PRIVKEY"],
+                    owner="mozilla-releng",
+                    repositories=["fxci-config"],
+                )
+            else:
+                client_cls = PublicClient
```
ahal, line 92: It might be nice to add a `simple_github.from_env` helper function that does this

PR #401, opened 2025-05-26
feat: use simple-github's client_from_env helper

bhearsum:


ahal (approved):
Thanks for doing this!

PR #405, opened 2025-05-29
chore: add alpha translations gpu pool

matt-boris:
Creating in order to test https://github.com/mozilla-platform-ops/worker-images/pull/314 in hopes of helping https://github.com/mozilla/translations/issues/1117.


bhearsum (approved):
This should be fine for the current need, where we're only running bicleaner. When we run the full pipeline we'll need to either revert back to the non-alpha workers or add alpha versions for the rest of the GPU workers.

PR #408, opened 2025-06-03
Add the required scopes and hooks for firefox ios to run automated betas

Eijebong:
This is a basically 8aceceb2d43a04a492d4adfd8263d4eb362d4df2, 22790145b650633e6c4e4d24c9e4569309fee097 and
0239d534afe83704fc1adb1e45d95e9b75546469 but for firefox-ios instead of staging-firefox-ios and targeting shipit prod


jcristau (approved):
Shouldn't we also have a prod level 1 client?

Eijebong:
> Shouldn't we also have a prod level 1 client?

Can't hurt I guess

PR #416, opened 2025-06-10
feat: add L1 production beetmover client for translations

bhearsum:


clients-interpreted.yml
```diff
@@ -150,6 +150,19 @@
...
+# match the block above
+- client:
+      project/releng/scriptworker/v2/beetmover/prod/firefoxci-{trust_domain}-1:
+          description: "{trust_domain} level 1 beetmover scriptworker"
+          scopes:
+              - queue:claim-work:scriptworker-k8s/{trust_domain}-1-beetmover
+              - queue:worker-id:{trust_domain}-1-beetmover/{trust_domain}-1-beetmover-*
+  for:
+      - project:
+            alias: [translations, staging-firefox-translations-training]
```
jcristau: should this be `{level: 1, feature: [beetmover]}`, instead of listing the projects?

jcristau (approved):
If we wanted to reduce duplication we could even remove the level here and drop the l1 client from the previous block for l3 projects.  But this seems fine either way.

PR #413, opened 2025-06-06
Remove $eval in $if condition for code-review testing hook

La0:
Followup #403

The [code-review-testing hook](https://firefox-ci-tc.services.mozilla.com/hooks/project-relman/code-review-testing) has been updated with the changes from #403, but all the trigerred tasks now fail with that error:

```
    TemplateError at template.payload.env[2]: $if can evaluate string expressions only
```

PR #403, opened 2025-05-26
Support Autoland or Mozilla-Central pulse trigger for code-review testing hook

La0:
Refs https://github.com/mozilla/code-review/pull/2801 & https://github.com/mozilla/code-review/issues/2773

The code-review hook must support pulse message triggers sent at the end of Mozilla Central & Autoland build groups.

We already did that by using the [events sub-system](https://github.com/mozilla/code-review/blob/95b29a76006f7c5462a7c096a95055b3b978ed33/events/code_review_events/workflow.py#L404), but now we only use the bot through its hook.

The hook will now have 2 pulse triggers:
1. from try pushes, transforming the payload into `TRY_RUN_ID`, `TRY_TASK_GROUP_ID` and `TRY_TASK_ID` environment variables
2. from Autoland / MC groups, transforming the payload into a unique `GENERIC_TASK_GROUP_ID`

## Sample payload from try push

```json
{
  "status": {
    "taskId": "ALywAOE0SDiwzjQMSkmTMQ",
    "provisionerId": "gecko-1",
    "workerType": "b-linux-gcp",
    "taskQueueId": "gecko-1/b-linux-gcp",
    "schedulerId": "gecko-level-1",
    "projectId": "none",
    "taskGroupId": "GXjaWu_kRoy6AP0yeeN60w",
    "deadline": "2025-05-27T15:09:57.208Z",
    "expires": "2025-06-23T15:09:57.208Z",
    "retriesLeft": 5,
    "state": "completed",
    "runs": [
      {
        "runId": 0,
        "state": "completed",
        "reasonCreated": "scheduled",
        "reasonResolved": "completed",
        "workerGroup": "us-central1-f",
        "workerId": "3186105317030741419",
        "takenUntil": "2025-05-26T15:40:28.694Z",
        "scheduled": "2025-05-26T15:20:26.826Z",
        "started": "2025-05-26T15:20:28.697Z",
        "resolved": "2025-05-26T15:20:40.971Z"
      }
    ]
  },
  "runId": 0,
  "task": {
    "tags": {
      "os": "linux",
      "kind": "code-review",
      "label": "code-review-issues",
      "project": "try",
      "retrigger": "false",
      "trust-domain": "gecko",
      "createdForUser": "reviewbot@noreply.mozilla.org",
      "worker-implementation": "docker-worker"
    }
  },
  "workerGroup": "us-central1-f",
  "workerId": "3186105317030741419",
  "version": 1
}
```

## from Autoland / MC groups

```json
{
[...]

PR #420, opened 2025-06-12
Support Autoland or Mozilla-Central pulse trigger for code-review production hook

La0:
This is the same patch as #403 + #413 for code-review production hook.

---
Refs https://github.com/mozilla/code-review/pull/2801 & https://github.com/mozilla/code-review/issues/2773

The code-review hook must support pulse message triggers sent at the end of Mozilla Central & Autoland build groups.

We already did that by using the [events sub-system](https://github.com/mozilla/code-review/blob/95b29a76006f7c5462a7c096a95055b3b978ed33/events/code_review_events/workflow.py#L404), but now we only use the bot through its hook.

The hook will now have 2 pulse triggers:
1. from try pushes, transforming the payload into `TRY_RUN_ID`, `TRY_TASK_GROUP_ID` and `TRY_TASK_ID` environment variables
2. from Autoland / MC groups, transforming the payload into a unique `GENERIC_TASK_GROUP_ID`

### DIFF
--- a/hooks/project-relman/code-review-production.yml
+++ b/hooks/project-relman/code-review-production.yml
@@ -39,12 +39,21 @@ payload:
       - $if: firedBy == 'pulseMessage'
         else: {}
         then:
-          TRY_RUN_ID:
-            $eval: payload.runId
-          TRY_TASK_GROUP_ID:
-            $eval: payload.status.taskGroupId
-          TRY_TASK_ID:
-            $eval: payload.status.taskId
+          $if: '"runId" in payload && "status" in payload'
+          then:
+            # Triggered by code-review task in try ending
+            # so we can analyze a try patch
+            TRY_RUN_ID:
+              $eval: payload.runId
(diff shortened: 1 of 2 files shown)


hooks/project-relman/code-review-production.yml
```diff
@@ -39,12 +39,21 @@ payload:
...
+            # so we can analyze a try patch
+            TRY_RUN_ID:
+              $eval: payload.runId
+            TRY_TASK_GROUP_ID:
+              $eval: payload.status.taskGroupId
+            TRY_TASK_ID:
+              $eval: payload.status.taskId
+
+          else:
+            # Triggered by end of build on autoland & MC
```
jcristau, line 54: I'm confused by this comment.  What about other l3 branches besides autoland and central?

La0: Other branches would trigger it as well, they are skipped in the bot's code 

I can update the comment to mention that.

PR #436, opened 2025-06-19
feat: don't fail cron runs when pushlog is empty

bhearsum:
This seems to be a footgun when setting up a new repository: you can't add it to fxci-config until you push to it, which is both unintuitive, and seems less than ideal (any push you make wouldn't have jobs run, i think?).

Theoretically this could happen if there's an issue on hg.m.o, but I'm not sure that's worth worrying about.

### DIFF
--- a/build-decision/src/build_decision/cron/__init__.py
+++ b/build-decision/src/build_decision/cron/__init__.py
@@ -9,6 +9,7 @@ from pathlib import Path
 
 from requests.exceptions import HTTPError
 
+from ..repository import NoPushesError
 from ..util.keyed_by import evaluate_keyed_by
 from ..util.schema import Schema
 from . import action, decision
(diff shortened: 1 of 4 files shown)


jcristau:
We have similar intermittent issues with hg.m.o (https://bugzilla.mozilla.org/show_bug.cgi?id=1961220 e.g.), that break the hg-push hooks, so I'm not sure it's not worth worrying about.  But I'll have to double check if any of the failure modes we've seen would be affected by this patch.

ahal (commented):
If a repo doesn't have pushes, it doesn't have a `.taskcluster.yml` and tc-admin will fail so we can't add it to `projects.yml` anyway. Though that's something we should / could *also* fix.

bhearsum:
> If a repo doesn't have pushes, it doesn't have a `.taskcluster.yml`. Therefore tc-admin will fail and we can't add it to `projects.yml` anyway. Though that's something we should / could _also_ fix.

That's not necessarily true though, I think? In the ESR case, the repo is cloned on the server side from an existing one, so it has a bunch of contents (including `.taskcluster.yml`), but no pushes.

PR #443, opened 2025-06-24
feat: Bug 1962119 - Remove experimental AMD pools and add c3d pool

hneiva:


worker-pools.yml
```diff
@@ -250,82 +250,50 @@ pools:
             - *scratch-disk
             - *scratch-disk
           machine_type: c2-standard-16
-  - pool_id: 'gecko-1/b-linux-gcp-bug1962119-c2d'
+  - pool_id: '{pool-group}/b-linux-gcp-c3d-16'
```
ahal, line 253: Typically we've avoided putting the instance name in the pools and instead use suffixes like -large, -xlarge, etc. This gives us some leeway to tweak the instance types without needing to update all the repos that use the pools.

Would it be bad to update existing `b-linux-gcp` pools in-place? If we need a separate pool, maybe `b-linux-amd` would be a better name (I'd also like to remove the `-gcp` suffix while we're at it)

ahal (changes requested):

worker-pools.yml
bhearsum: If we really want to future proof things, adding a version suffix or infix might work (as opposed to inventing/reinventing things like `gcp`, `amd`, `d2g`, etc.).

ahal: Yeah, that might be better.. though it imparts less information about the pool and still requires us to go update every consumer.

Maybe we need aliases like the images have. So we can just stick the entire instance type into the pool name, but then also have simplified aliases that we can point at whatever. Then consumers could choose to "pin" to a specific instance type, or they could use the alias and trust us to use what's best.

I think in this case swapping in-place might make the most sense? I don't know why we'd need both intel and amd versions of a pool, and if we've already established amd is cheaper, why not switch everything?

bhearsum: The main benefit to having new pools is that when we swap things in, they show up on treeherder, and bisectable. I'm not expert enough here to know if this change is safe enough to just swap in everywhere though.

PR #445, opened 2025-06-24
Bug 1965145 - Add new lando scopes for Thunderbird repos

dandarnell:
Adding these to hopefully fix our merge automation.

I've excluded the android scopes, since Thunderbird for Android uses our own infrastructure.


wsmwk:
@jcristau can you review?

PR #448, opened 2025-06-26
Feat: Remove remaining experimental pools and add large/xlarge c3d pools

hneiva:


src/ciadmin/generate/worker_pools.py
```diff
@@ -609,6 +609,13 @@ def update_config(config, name, attributes):
             else:
                 del config["worker-manager-config"]["launchConfigId"]
 
+        if attributes.get("scratch-disks", 0) > 0:
+            scratch_disks = [attributes["scratch-disk-definition"]] * attributes[
+                "scratch-disks"
+            ]
+            for instance in config.get("instance_types", []):
+                instance["disks"] = instance.get("disks", []) + scratch_disks
+
```
hneiva, line 618: TBH I'm not super happy with this approach to add scratch disks per variant, but couldn't think of a better way

worker-pools.yml
```diff
@@ -250,25 +250,113 @@ pools:
...
             - *scratch-disk
           machine_type: c2-standard-16
-  - pool_id: '{pool-group}/b-linux-amd'
+  - pool_id: '{pool-group}/b-linux{suffix}-amd'
     description: Worker for Firefox automation.
     owner: release+tc-workers@mozilla.com
     email_on_error: true
+    attributes:
+      # Used to populate scratch disks
+      scratch-disk-definition: *scratch-disk
```
ahal, line 259: Maybe eventually we should move the `*scratch-disk` config into code, and just make every pool set `num_scratch_disks` or something like that.

worker-pools.yml
```diff
@@ -250,25 +272,91 @@ pools:
...
     variants:
+      # Base 16-core (gecko, comm, app-services, mozillaonline)
       - pool-group: gecko-1
+        suffix: ''
+        instance_types: [*c3d-standard-16-lssd]
```
ahal, line 286: Nice! I like this as a compromise until we sort this out more robustly later.

PR #466, opened 2025-07-21
Remove grants on secrets that are not used anymore

Eijebong:


Eijebong:
Let's wait until the next release before merging this and clearing out the secrets, just in case we need to revert

jcristau (approved):

PR #468, opened 2025-07-23
ci: allow manually triggering the 'deploy' action

ahal:
I think this is all that's needed...

But it's hard to test because it needs to exist on the default branch first.


ahal:
I'll add that we can apparently associate this to a Github environment to lock it down further.. but given this workflow runs on push, anyone with write access can trigger it at anytime via pushing commits anyway.

PR #475, opened 2025-07-25
Support rendering .tc.yml with cron input for decision type

ahal:
When triggering a cron hook, it's possible to specify a payload at trigger time. For the "action" type of cron task, it's further possible to forward this input to the action as action input. But for the decision cron type, this input is dropped.

This change adds a way for the input to get added to the context used to render the .taskcluster.yml file, which allows consumers to use it with the "decision" type of cron.

### DIFF
--- a/build-decision/src/build_decision/cron/decision.py
+++ b/build-decision/src/build_decision/cron/decision.py
@@ -4,6 +4,7 @@
 
 
 import copy
+import json
 import os
 import shlex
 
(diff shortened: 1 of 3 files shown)


build-decision/src/build_decision/cron/decision.py
```diff
@@ -39,12 +40,20 @@ def run_decision_task(job_name, job, *, repository, push_info, dry_run):
 
     arguments = make_arguments(job)
 
+    cron_input = {}
+    if job.get("include-cron-input") and "HOOK_PAYLOAD" in os.environ:
```
ahal, line 44: I added `include-cron-input` to make it consistent with the "action" type, but I'm wondering if we shouldn't just always forward "HOOK_PAYLOAD" to the render context. I can't really think of any risks with doing that

ahal: Maybe I'll just leave it in-place for now to be consistent (and out of an abundance of caution). I'll file an issue to look into removing both of them later.

PR #481, opened 2025-07-30
Run crontrigger hook every 3 hours instead of 30 minutes

marco-c:


jcristau (approved):

hooks.yml
```diff
@@ -86,7 +86,7 @@ project-relman/code-coverage-crontrigger-production:
   owner: release-mgmt-analysis@mozilla.com
   email_on_error: true
   schedule:
-    - 0 */30 * * * *
+    - 0 0 */3 * * *
```
JohanLorenzo, line 89: Note to self: this syntax slightly differs from https://cron.help. Per https://www.npmjs.com/package/cron-parser, the first field is for the seconds. cron.help doesn't have seconds there. 

Thus, I confirm it's going to schedule the cron every 3 hours starting from midnight.

PR #483, opened 2025-08-05
add publicIP to all alpha windows workers

jwmoss:
This PR enables `publicIp: true` on all windows alpha pools, while setting it to `publicIp: false` on all other pools. Additionally, this adds functionality to the worker-manager-config section to support `publicIp` property with the keyed-by pattern `by-pool-id`.

### DIFF
--- a/src/ciadmin/generate/worker_pools.py
+++ b/src/ciadmin/generate/worker_pools.py
@@ -337,16 +337,17 @@ def get_azure_provider_config(
                 attrs,
             )
 
-            public_ip_defined = (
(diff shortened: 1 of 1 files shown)


ahal (changes requested):
So doing a bit of research, this seems like at a minimum we'll need an RRA. We'll need to make sure we have firewalls / WAF setup properly.

It's also possible that we should just eat the extra cost here to avoid assigning public IPs.

But for now, this is not a decision that can be made in a pull request.

PR #490, opened 2025-08-14
feat: add xlarge pool for t-linux-docker

bhearsum:
This is needed to run `tgdiff` without OOMing after adding multiprocess support.


hneiva (approved):
I'm moving a bunch of tests to AMD hardware, if you can test with that as well it would be fantastic

PR #491, opened 2025-08-15
Bug 1983581 - Create new AMD pools for t-linux-docker

hneiva:


hneiva:
Self note: Only merge this after soft freeze

PR #496, opened 2025-08-19
Ensure regions are sorted by both tc-admin and fxci-config for clean diffs

ahal:


ahal:
This change results in an empty tc-admin diff from current main.

PR #500, opened 2025-08-20
Prevent grants that reference non-existent pools

ahal:


src/ciadmin/check/check_grants.py
```diff
@@ -44,6 +44,88 @@ async def check_grant_aliases():
...
+    # check is to keep grants.yml clean, not the generated grants.
+    grants = await Grant.fetch_all()
+    pools = [p.workerPoolId for p in generated.filter("WorkerPool=.*")]
+    invalid_scopes = set()
+
+    for grant in grants:
+        for scope in grant.scopes:
+            assert isinstance(scope, str)
+
+            if not any(scope.startswith(p) for p in prefixes):
```
jcristau, line 87:
```suggestion
            if not scope.startswith(tuple(prefixes)):
```
(or make prefixes a tuple and `...startswith(prefixes)`)

jcristau (commented):
I like this, thanks!

src/ciadmin/check/check_grants.py
ahal: I always forget that `startswith` can do that.

jcristau (approved):

PR #504, opened 2025-08-22
Bug 1984295 - add nss pool for win11 arm64 tester/builder

jwmoss:


scoutzeng:
Hi @jwmoss ,  I'm curious about what host is used to compile NSS for Windows ARM64 — is it Linux on ARM64 or some other combination?
Is the build.sh script provided in the [NSS code repo](https://hg.mozilla.org/projects/nss) for this? I looked through the contents of **build.sh**, but it doesn't seem to support cross-compilation or explicitly target Windows on ARM64. Neither does Makefile.

PR #506, opened 2025-08-23
Bug 1984773 - Enable more zones in GCP

hneiva:
There are unused zones within regions we already have pools.

Some compute types (or GPUs) are not available in some zones - restrictions are listed under `invalid-instances`.

Availability links:
GPUs: https://cloud.google.com/compute/docs/gpus/gpu-regions-zones#view-using-table Machine types: https://cloud.google.com/compute/docs/regions-zones#available


ahal (approved):

hneiva:
perfect, it's failing where I expect it to fail.
I'll fix those pools before merging this

hneiva:
#509 fixes the failing tests

tests/ciadmin/test_generate_worker_pools.py
```diff
@@ -293,3 +297,115 @@ async def test_make_worker_pool(
     # Assertion callbacks correspond to `id` of parametrized test.
     param_id = request.node.callspec.id
     globals()[f"assert_{param_id}"](result)
+
+
+@pytest.mark.asyncio
+async def test_google_machine_types_available(environment, mocker):
```
ahal, line 303: I think this should be a check rather than a unittest. These tests are more for testing the actual logic of `ciadmin`, e.g using fake data. Whereas the checks are more for testing the configuration itself.

A good example to follow would be:
https://github.com/mozilla-releng/fxci-config/blob/52de9c364ac200bb66be863d0c024244ef4d486e/src/ciadmin/check/check_worker_pools.py#L152

ahal (changes requested):

PR #514, opened 2025-08-28
chore: Refactor gcp machine checks + add zone check

hneiva:


ahal (approved):
Nice, lgtm! Does this run relatively quickly? Just asking because it'll add a bit of time to the deploy pipeline.

hneiva:
Yeah should only take a few seconds

PR #516, opened 2025-08-28
fix: use globally shared Github client

ahal:
Previously we were instantiating two SimpleGithub clients for each project. One to get the branches, another to get the .taskcluster.yml.

On my system, creating this many clients (which in turn each create their own aiohttp sessions and DNSResolvers), was triggering a bug in aiodns by surpassing some sort of resource limit:
https://github.com/aio-libs/aiodns/issues/128

I'm not really sure why no one else was hitting this, but using a single shared `SimpleGithub` client across all Github API calls fixes the issue for me and should be faster anyway as we will actually re-use aiohttp sessions.

### DIFF
--- a/src/ciadmin/util/github.py
+++ b/src/ciadmin/util/github.py
@@ -1,0 +1,38 @@
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at http://mozilla.org/MPL/2.0/.
+
+import asyncio
+
+from simple_github import AsyncClient, client_from_env
+
+# Global shared client session
+_client: AsyncClient | None = None
+_client_lock = asyncio.Lock()
+
+
+async def get_client():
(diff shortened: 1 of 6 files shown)


src/ciadmin/util/github.py
```diff
@@ -0,0 +1,37 @@
...
+
+    return _client
+
+
+async def close_client():
+    """Cleanup the shared global client."""
+    global _client
+
+    if _client is not None:
+        await _client.close()
```
jcristau, line 36: Doesn't this need to take the client_lock?

src/ciadmin/util/github.py
```diff
@@ -0,0 +1,38 @@
...
+
+    return _client
+
+
+async def close_client():
+    """Cleanup the shared global client."""
+    global _client
+
+    if _client is not None:
+        async with _client_lock:
```
jcristau, line 36: Swap those two otherwise another thread can set _client to None between the check and taking the lock and the next line would crash.

PR #528, opened 2025-09-04
refactor: move translations scopes to their own file

bhearsum:
This is minor follow-up to #524 to move at least a few scopes that are obviously grouped together to a dedicated file.


ahal (approved):
Thanks!

PR #533, opened 2025-09-10
chore(deps): add pre-commit

jwmoss:
When running `uv run tc-admin diff --environment firefoxci 2>&1 | tee diff.out` for https://github.com/mozilla-releng/fxci-config/pull/504 locally I was unable to perform a git commit without installing pre-commit. Not sure if this was intentional or not.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,7 @@ dev = [
     "flake8",
     "google-cloud-compute",
     "pip>=21.2",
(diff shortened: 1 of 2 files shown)


ahal (approved):
Seems reasonable. I guess we mostly have it installed globally outside of any specific repositories, but I see no downside to including it in the project's dev dependencies.

PR #329, opened 2025-03-27
Allow overrides for maxCapacity and initialWeight per launch config

lotas:
Worker pool's top level `worker-manager-config` allows specifying `maxCapacity` and `initialWeight` with `by-xx` functionality, which makes it possible to set values based on region/zone/machine type.

If `worker-manager-config` is given for `vmSizes` (Azure) or `instance_types` (GCP), it will be passed as is to the resulting config.

Example usage:

  ```yml
  - pool_id: w/p
    ..
    config:
      ..
      vmSizes:
        - vmSize: Standard_F32s_v2
          ..
          worker-manager-config:
            # Option1: top priority for resulting config
            maxCapacity: 10
            initialWeight: 0.5
      worker-manager-config:
        capacityPerInstance: 1
        # Option2: lower priority and possibility to specify by region and vm size
        maxCapacity:
          by-location:
            central-us: 10
            west-us-2: 5
            east-us-2: 5
            default: 20
        initialWeight:
          by-vmSize:
            default: 0.88
            Standard_F32s_v2: 0.7
        # Option3: combination of attributes
        initialWeight:
          by-vmSize:
            default: 1
            Standard_.*:
              by-location:
                west-us-2: 0.8
                east-us-1: 0.8
                default: 1
```

Supported attributes for `Azure`:
- `location`
- `vmSize`

Supported attributes for `GCP`:
- `zone`
- `region`
- `machineType`

Supported attributes for `AWS`:
- `availabilityZone`
- `instanceType`

PR #539, opened 2025-09-18
chore(launchconfig): order weights by spot cpu price per region per vm size

jwmoss:
Similar to https://github.com/mozilla-releng/fxci-config/pull/329


lotas (approved):
LGTM. Aligns with what was added in #329

ahal (approved):

PR #543, opened 2025-09-22
Bug 1989066 - check that trusted generic-worker pools don't allow privileged tasks

jcristau:


src/ciadmin/check/check_privileged.py
```diff
@@ -36,6 +38,35 @@ async def check_privileged_is_untrusted(generate_resources):
...
+                if (
+                    gwConfig.get("config", {})
+                    .get("d2gConfig", {})
+                    .get("enableD2G", False)
+                ):
+                    d2gConfig = gwConfig["config"]["d2gConfig"]
+                    if d2gConfig.get("allowPrivileged", True):
+                        privileged = True
+                        break
+                    if d2gConfig.get("allowGPUs", True):
```
matt-boris, line 58: `allowGPUs` defaults to `False`

Should be:


```suggestion
                    if d2gConfig.get("allowGPUs", False):
```

matt-boris: Also, why is this considered privileged, just wondering?

matt-boris (changes requested):

src/ciadmin/check/check_privileged.py
jcristau: While the default is currently False in g-w, I think it's safer / more future-proof to explicitly disable it here for trusted pools.

As to why I'd consider it privileged, gpus add a massive attack surface so at this stage I'd prefer to blanket disallow them for trusted pools at least until such time as there's a use case and we need to revisit.

matt-boris (approved):
LGTM, thank you 👍🏻

src/ciadmin/check/check_privileged.py
```diff
@@ -13,15 +13,22 @@
 @pytest.mark.asyncio
 async def check_privileged_is_untrusted(generate_resources):
     """
-    Ensures that any docker-worker pool with `allowPrivileged` is not
-    run on a trusted image.
+    Ensures that trusted pools (running on a trusted image or trusted provider)
+    have privileged config options turned off:
+
+    - for docker-worker, ensure that `allowPrivileged` is disabled
+    - for generic-worker, ensure that `enableRunTaskAsCurrentUser` is disabled
```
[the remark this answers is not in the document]
matt-boris: He sets it to false in this PR too :)

https://github.com/mozilla-releng/fxci-config/pull/543/files#diff-5b0199d199a0681aeb1e93b70e433a60fc3b4c67e8c0483b122b8ba7e10f5741

PR #548, opened 2025-09-22
feat: add new test kinds to os-integration tests (bug 1989652)

bhearsum:
Only `browsertime` actually has os-integration tests at the moment, but AFAICT there's no harm in futureproofing this.


bhearsum:
This goes with the stack in https://bugzilla.mozilla.org/show_bug.cgi?id=1989652, but I think it's safe to land it anytime?

bhearsum:
> No Taskcluster jobs started for this command

this is expected; the fact that it ran green is all I was looking for

PR #556, opened 2025-09-24
Bug 1773091 - add necessary beetmover scopes for autoland

hneiva:


bhearsum (commented):
afaict the check failures are because you're enabling the `beetmover` feature, which ends up adding the clients from `clients-interpreted.yml`, which conflict with the hardcoded gecko clients in `clients.yml`.

I don't think you need any changes to scopes: autoland already has `[queue:create-task:highest:scriptworker-k8s/gecko-3-*](https://firefox-ci-tc.services.mozilla.com/auth/scopes/queue%3Acreate-task%3Ahighest%3Ascriptworker-k8s%2Fgecko-3-*)`, which is what it should need to create these tasks. If they're still trying to create `gecko-1` tasks in https://phabricator.services.mozilla.com/D218468, you need to address that there.

hneiva:
I've switched it to explicitly give the scopes to autoland

PR #567, opened 2025-10-02
feat: add northamerica-northeast1 region for b-linux-medium-gcp pool

bhearsum:
It looks like we largely don't use this region due to https://bugzilla.mozilla.org/show_bug.cgi?id=1882320, which says:
> This somewhat stalled out because there are certain instance types that are significantly more expensive in this region. We need some kind of way to control which instance types are allowed in which regions. Maybe just a check in fxci-config is good enough.

The `c2-standard-8` used for this worker type is actually cheaper in this region though...so this concern doesn't apply.

### DIFF
--- a/worker-pools.yml
+++ b/worker-pools.yml
@@ -862,7 +862,7 @@ pools:
           gecko-2: 50
           gecko-3: 100
           default: 10
-      regions: [us-central1, us-west1]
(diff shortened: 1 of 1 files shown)


bhearsum:
We do! See https://bugzilla.mozilla.org/show_bug.cgi?id=1868861

PR #899, opened 2026-03-24
feat: migrate fuzzing project from community-tc to fxci

petemoore:
## Summary

Migrate all fuzzing project entities from `community-tc-config` to `fxci-config`, enabling the fuzzing team's workloads to run on the firefox-ci Taskcluster deployment.

### What's included

- **15 worker pools** (GCP, AWS, Azure) added to `worker-pools.yml`
- **11 hooks** with template files in `hooks/project-fuzzing/`
- **10 grant entries** producing 22 new roles in `grants.yml`
- **6 worker image entries** in `worker-images.yml`
- **`externally-managed.yml`** — new config file + Python support to prevent `tc-admin` from deleting resources in the `proj-fuzzing`/`project-fuzzing` namespaces that are deployed externally by `fuzzing-tc-config`. This mirrors `externallyManaged: true` behavior in `community-tc-config`.
- **`generate_subset.py`** — dev tool to run `tc-admin generate` for specific resource types without needing `GITHUB_TOKEN` (works around the `--resources` flag parsing bug in `boot.py`)

### Provider mapping

| community-tc | fxci |
|---|---|
| `community-tc-workers-google` | `fxci-level1-gcp` |
| `community-tc-workers-aws` | `aws` |
| `community-tc-workers-azure` | `azure2` |

### How it was validated

1. Generated baseline JSON snapshots from both repos using `tc-admin generate --json`
2. Applied the changes to fxci-config
3. Generated a new fxci snapshot
4. Compared: all 984 pre-existing fxci resources preserved, 48 new fuzzing resources added
5. Cross-referenced against community-tc's 56 fuzzing resources — 40 directly match, 16 are expected gaps (worker-pool roles, docker-worker secrets, and an externally-managed hook role that differ architecturally between the repos)

Full validation results: `docs/fuzzing-migration-validation.md`

Reproducible via:
```bash
python generate_subset.py --environment=firefoxci --json > /tmp/fxci-after.json
python tests/test_fuzzing_migration.py \
    --community-tc-json /tmp/community-tc-all.json \
    --fxci-before-json /tmp/fxci-before.json \
    --fxci-after-json /tmp/fxci-after.json
```

### Known limitations / follow-up items

- Azure and AWS images need to be shared from community-tc cloud accounts to fxci accounts before the pools actually work in production
- The `failed-log-ingestor` hook (externally managed by fuzzing-tc-config) is not included — it will be preserved by the externally-managed exclusion patterns

## Test plan

- [ ] Verify `tc-admin generate` produces expected output locally
- [ ] Verify `tc-admin diff` against staging shows only additions (no deletions of existing resources)
- [ ] Verify fuzzing-tc-config's externally-deployed resources are not affected
- [ ] Coordinate with RelOps team on image sharing between cloud accounts

🤖 Generated with [Claude Code](https://claude.com/claude-code)


docs/fuzzing-migration-validation.md
```diff
@@ -0,0 +1,86 @@
+# Fuzzing Migration Validation Results
```
ahal: I'd prefer to not commit this. If there's any info worth preserving I suggest we add it to the commit message.

grants.d/grants.yml
```diff
@@ -2215,6 +2215,124 @@
...
+  to:
+    - roles:
+        - github-team:MozillaSecurity/tc-admin
+
+# Worker pool roles for fuzzing pools (matches community-tc behavior)
+- grant:
+    - auth:sentry:generic-worker
+  to:
+    - roles:
+        - worker-pool:proj-fuzzing/*
```
ahal: I think my `grants/<trust-domain>.yml` files came after this PR, but can you move this to a `grants/fuzzing.yml` file? It'll get picked up automatically.

grants.d/grants.yml
```diff
@@ -2215,6 +2215,124 @@
...
+    - queue:create-task:highest:proj-fuzzing/ci-*
+    - queue:create-task:highest:proj-fuzzing/decision
+    - queue:route:checks
+    - secrets:get:project/fuzzing/deploy-*
+    - secrets:get:project/fuzzing/codecov-*
+    - secrets:get:project/fuzzing/pypi-*
+    - secrets:get:project/fuzzing/ci-*
+  to:
+    - roles:
+        - repo:github.com/MozillaSecurity/*
```
ahal: This will grant access to these secrets to any pull request on any repo in the org. I'm assuming we don't want that unless these secrets are all public anyway?

grants.d/grants.yml
```diff
@@ -2215,6 +2215,124 @@
...
+- grant:
+    - queue:scheduler-id:fuzzing
+  to:
+    - roles:
+        - hook-id:project-fuzzing/*
+        - repo:github.com/MozillaSecurity/*
+
+- grant:
+    - queue:create-task:highest:proj-fuzzing/ci
+    - queue:create-task:highest:proj-fuzzing/ci-*
```
ahal: Is there a concept of level here? If so this might be granting L3 queues to PRs.

grants.d/grants.yml
```diff
@@ -2215,6 +2215,124 @@
...
+    - queue:scheduler-id:fuzzing
+    - queue:cancel-task:fuzzing/*
+    - queue:route:notify.*
+    - secrets:get:project/fuzzing/*
+    - worker-manager:manage-worker-pool:proj-fuzzing/*
+    - worker-manager:provider:aws
+    - worker-manager:provider:azure2
+    - worker-manager:provider:fxci-level1-gcp
+  to:
+    - roles:
```
ahal: Granting to roles seems fine for a first pass, but we'll likely want to use `fxci-config`'s richer "project" / "job" grants eventually.

grants.d/grants.yml
```diff
@@ -2215,6 +2215,124 @@
...
+        - repo:github.com/MozillaSecurity/fuzzmanager-config:*
+
+- grant:
+    - queue:get-artifact:project/fuzzing/private/logs/*
+  to:
+    - roles:
+        - login-identity:github/173702880|moz-mdauer
+
+- grant:
+    - docker-worker:capability:privileged
```
ahal: This grants root access to the worker to PRs.

grants.d/grants.yml
```diff
@@ -2215,6 +2215,124 @@
...
+    - index:insert-task:project.fuzzing.orion.*
+    - queue:route:index.project.fuzzing.orion.*
+    - secrets:get:project/fuzzing/docker-hub
+  to:
+    - roles:
+        - repo:github.com/MozillaSecurity/orion:*
+
+- grant:
+    - docker-worker:capability:privileged
```
ahal: Ditto

src/ciadmin/generate/ciconfig/externally_managed.py
```diff
@@ -0,0 +1,59 @@
...
+            relevant_exclusions.append(pat)
+
+    if not relevant_exclusions:
+        resources.manage(base_pattern)
+        return
+
+    # Build a negative lookahead pattern
+    exclusion_alts = "|".join(relevant_exclusions)
+    pattern = f"(?!{exclusion_alts}){base_pattern}"
+    resources.manage(pattern)
```
ahal: I'm not sure it's better or not, but here's a suggestion.

What if we manage everything but then call `Resources.filter` to get a new Resources object with the exclusions applied?
https://github.com/taskcluster/tc-admin/blob/main/tcadmin/resources/resources.py#L153

Feels like it might be a bit simpler.

generate_subset.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python3
+"""
+Generate a subset of fxci-config resources without needing GITHUB_TOKEN.
+Skips in_tree_actions, cron_tasks, hg_pushes, and scm_group_roles which
+need network access or are not relevant to the fuzzing migration.
+
+Usage:
+    python generate_subset.py --environment=firefoxci --json > output.json
+    python generate_subset.py --environment=firefoxci --text
+"""
```
ahal: I'm a bit confused by this file, was this some debug script that accidentally got committed?

worker-pools.yml
```diff
@@ -4557,3 +4557,273 @@ pools:
...
+      maxCapacity: 10
+      implementation: generic-worker/worker-runner-linux-multi
+      regions: [us-central1, us-west1]
+      image: fuzzing-gw-ubuntu-24-04
+      instance_types:
+        - disks:
+            - <<: *persistent-disk
+              diskSizeGb: 60
+            - *scratch-disk
+          machine_type: n2-standard-4
```
ahal: It doesn't necessarily need to block this PR I guess, but there's a lot of duplication in these pools. We should use the yaml anchors / pool-groups feature that we use other places. Can probably reduce these to one or two stanzas.

ahal (changes requested):

hooks.yml
```diff
@@ -227,3 +227,268 @@ project-engwf/reviewer-assignment-infra-testing:
...
+    - queue:create-task:highest:proj-fuzzing/bugmon-processor
+    - queue:create-task:highest:proj-fuzzing/bugmon-processor-windows
+    - queue:get-artifact:project/fuzzing/bugmon/*
+    - queue:route:notify.email.jkratzer@mozilla.com.on-failed
+    - queue:scheduler-id:fuzzing
+    - secrets:get:project/fuzzing/bz-api-key
+    - secrets:get:project/fuzzing/pernosco-group
+    - secrets:get:project/fuzzing/pernosco-secret
+    - secrets:get:project/fuzzing/pernosco-user
+  template_file: hooks/project-fuzzing/bugmon.yml
```
ahal: The template dir changed names `hooks` -> `hook-templates`

hooks.yml
```diff
@@ -227,3 +227,268 @@ project-engwf/reviewer-assignment-infra-testing:
...
+    - queue:create-task:highest:proj-fuzzing/bugmon-processor-windows
+    - queue:get-artifact:project/fuzzing/bugmon/*
+    - queue:route:notify.email.jkratzer@mozilla.com.on-failed
+    - queue:scheduler-id:fuzzing
+    - secrets:get:project/fuzzing/bz-api-key
+    - secrets:get:project/fuzzing/pernosco-group
+    - secrets:get:project/fuzzing/pernosco-secret
+    - secrets:get:project/fuzzing/pernosco-user
+  template_file: hooks/project-fuzzing/bugmon-confirm.yml
+  trigger_schema: {}
```
ahal, line 431: There's a new "variants" feature that got implemented since you wrote this. It should allow you (or Claude) to greatly reduce the amount of duplication in these hooks.

It *could* be a follow-up, but Claude should be able to do a consolidation pretty trivially.

src/ciadmin/generate/ciconfig/externally_managed.py
```diff
@@ -0,0 +1,59 @@
...
+
+async def get_externally_managed_patterns():
+    """
+    Load externally-managed.yml and return a flat list of all regex pattern
+    strings across all projects.
+    """
+    data = await get_ciconfig_file("externally-managed.yml")
+    patterns = []
+    for project_patterns in data.values():
+        patterns.extend(project_patterns)
```
ahal: You can use `itertools.chain` for this

src/ciadmin/generate/grants.py
```diff
@@ -234,13 +238,14 @@ async def update_resources(resources):
     grants = await Grant.fetch_all()
     projects = await Project.fetch_all()
     environment = await Environment.current()
+    ext_patterns = await get_externally_managed_patterns()
```
ahal: Instead of importing this in every resource generation file and then passing the results into `manage_with _exclusions`, why not just have `manage_with_exclusions` call `get_externally_managed_patterns()` itself?

src/ciadmin/generate/ciconfig/externally_managed.py
```diff
@@ -0,0 +1,59 @@
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at http://mozilla.org/MPL/2.0/.
+
+import re
+
+from .get import get_ciconfig_file
+
+
+async def get_externally_managed_patterns():
```
ahal, line 16: We probably want to use `functools.cache` here

externally-managed.yml
```diff
@@ -0,0 +1,21 @@
+---
+# Resources matching these patterns are externally managed and should NOT be
+# deleted by tc-admin even if they are not generated by this repository.
+#
+# This is needed for projects like fuzzing where a separate repo
+# (fuzzing-tc-config) deploys additional hooks, worker pools, and roles
+# into the same namespace that this repo also partially manages.
+#
+# Each entry is a regex pattern matching tc-admin resource IDs.
+# Format: Kind=identifier (e.g., WorkerPool=proj-fuzzing/.*, Hook=project-fuzzing/.*)
```
ahal, line 10: I still don't love this, but understand it's a way to move the project forward. We (releng) will likely be pushing to remove the externally managed stuff over time. But that's not your concern :)

ahal (changes requested):

worker-pools.yml
```diff
@@ -4581,3 +4593,244 @@ pools:
...
+    email_on_error: false
+    provider_id: fxci-level1-gcp
+    config:
+      minCapacity: 0
+      maxCapacity: 10
+      implementation: generic-worker/worker-runner-linux-multi
+      regions: *default-gcp-regions
+      image: fuzzing-gw-ubuntu-24-04
+      instance_types:
+        - *n2-standard-4-60gb-lssd
```
ahal: Please also use pool variants to reduce the duplication here. Claude can help

src/ciadmin/generate/ciconfig/externally_managed.py
petemoore: I think `functools.cache` doesn't apply to coroutines (it would cache the awaitable, which can only be awaited once)?

worker-pools.yml
petemoore: who is claude?

petemoore:
Hi Andrew! I think I've addressed everything. I also realised I need to port a couple of parameterized roles across for repo-admin and project-admin so that fuzzing can continue to self-serve in their fuzzing-tc-config repo. None of this impacts any other roles or clients in fxci, so should be safe. Many thanks for your reviews!

src/ciadmin/generate/ciconfig/externally_managed.py
ahal: Ah true!

externally-managed.yml
```diff
@@ -0,0 +1,21 @@
...
+
+fuzzing:
```
ahal, line 12: Looks like the keys in this file are unused.. Not a requirement for landing, but I'd mildly prefer it just be a top-level list.

Or maybe an explicit:
```
resources:
  - ...
```

and have that be the only key we check.

ahal (approved):
Thanks, lgtm!

Before we land this, could you run a `tc-admin diff` and just double check there's nothing unexpected showing up?

grants.d/grants.yml
```diff
@@ -87,6 +87,70 @@
...
+    - worker-manager:manage-worker-pool:proj-<..>/*
+    - worker-manager:provider:aws
+    - worker-manager:provider:azure2
+    - worker-manager:provider:fxci-level1-gcp
+    - worker-manager:provider:null-provider
+    - worker-manager:remove-worker:proj-<..>/*
+  to:
+    - roles:
+        - project-admin:*
+
```
ahal, line 144: On second thought, can you remove these grants? We don't use `project-admin:` or `repo-admin:` and I don't really understand where they come from. Let's stick to only granting stuff to fuzzing related resources in this PR.

ahal (changes requested):

grants.d/grants.yml
```diff
@@ -87,6 +87,70 @@
...
+    - auth:delete-role:hook-id:project-<..>/*
+    - auth:delete-role:project:<..>:*
+    - auth:disable-client:project/<..>/*
+    - auth:enable-client:project/<..>/*
+    - auth:reset-access-token:project/<..>/*
+    - auth:update-client:project/<..>/*
+    - auth:update-role:hook-id:project-<..>/*
+    - auth:update-role:project:<..>:*
+    - docker-worker:cache:*
+    - generic-worker:cache:*
```
ahal: This doesn't look safe to grant to something that's supposed to be project scoped.

grants.d/grants.yml
```diff
@@ -87,6 +87,70 @@
...
+    - secrets:get:project/<..>/*
+    - secrets:get:worker-pool:proj-<..>/*
+    - secrets:set:project/<..>/*
+    - secrets:set:worker-pool:proj-<..>/*
+    - worker-manager:create-worker:proj-<..>/*
+    - worker-manager:manage-worker-pool:proj-<..>/*
+    - worker-manager:provider:aws
+    - worker-manager:provider:azure2
+    - worker-manager:provider:fxci-level1-gcp
+    - worker-manager:provider:null-provider
```
ahal: Ditto, I don't really understand what these grants allow.. but they aren't project scoped by the role.

hook-templates/project-fuzzing/grizzly-reduce-reset-error.yml
```diff
@@ -0,0 +1,38 @@
...
+    - '${payload.status.taskId}'
+    - '--auto'
+    - '--only-if-quality'
+    - '4'
+  artifacts:
+    project/fuzzing/private/logs:
+      path: /logs/
+      type: directory
+metadata:
+  name: grizzly-reduce-monitor
```
ahal, line 25: This has the wrong name

grants.d/grants.yml
petemoore: So these roles come from community, and the plan was to lift everything as-is (fuzzing is just the first project of several to migrate, and they all use this structure). This is the backbone of the current project organisation that the projects currently use and depend on for all their project grants, namespaces, etc. I think not including this and hardcoding it in fuzzing roles and clients would complicate matters when we come to move other projects later. Note, the tc-admin diff shows that this PR changes no existing resources whatsoever, and so no existing security boundaries are changed. The expanded scopes for all existing fxci entities remains unchanged. With these roles, we can comfortably lift all the current projects from community-tc and put them in fxci and be confident that they are segregated from each other and all fxci entities.

grants.d/grants.yml
petemoore: I agree this is not ideal, but again, this is just lifting the existing configuration from community-tc. We can file a separate issue for that at source, but I think it isn't related to the migration, it is an existing issue. Fortunately projects run on their own project worker pools, so cross-project pollution isn't possible with scopes, but this also means essentially any project task can affect any project cache for its own project.

grants.d/grants.yml
petemoore: As above, this is just to mirror community-tc existing grants. This is to have fidelity with the source where the entities are being migrated from.

hook-templates/project-fuzzing/grizzly-reduce-reset-error.yml
petemoore: Well spotted. Again, this is an issue at source, but we can probably reasonably safely fix it here as part of the migration as it is just a name.

grants.d/grants.yml
ahal: I understand that the roles are just being copy/pasted from community.. but by porting this over to fxci, you are giving these community projects access to mess with caches for any other project. You're right that they're using different worker pools, so in theory other caches shouldn't exist, but we do have many projects that share worker pools, and this is getting rid of a layer of protection.

E.g we take pains to scope caches by trust domain and level:
https://github.com/mozilla-releng/fxci-config/blob/fec69ae364722a27e668a7e715fa4bc9b289259c/grants.d/grants.yml#L950

Especially because fuzzing manages its own resources, I trust fuzzing much less than I trust any other project in fxci and I think it's important we sandbox them to just scopes that affect their own stuff.

I suggest we ask fuzzing to migrate their cache names to ones that follow convention I linked above. Maybe we can do this over on community as a pre-requisite and that way we can still preserve fidelity in the migration?

grants.d/grants.yml
ahal: Similar to the * caches, I need to understand what these grants allow before r+'ing. The fact is that this is giving scopes to community projects that might affect the overall instance. Maintaining security is more important than keeping 1 to 1 mapping of roles across the migration.

ahal (approved):
Thanks, this looks great!

jwmossmoz (changes requested):
@petemoore @ahal We either need to setup a new provider `azure-tceng` (which points to `Taskcluster Engineering DevTest Subscription`) as that's the location of the image. The other thing is we'll need to make sure that the resources in that azure subscription are configured in `azure-tceng` similar to how we do it today in `azure2`.

For aws, similarly, we'll need to setup a new provider `aws-tceng` and possibly provision networks that are configured. 

For gcp, I think the images will work as is.

petemoore:
I am happy to introduce new providers - that also works for me. In this PR i mapped community-tc providers to existing fxci providers, but if we have new ones, then that is totally doable too. Those providers could share the config with community-tc so indeed no duplication of images etc, they just take them from the same source, and when community-tc is decommissioned, the accounts continue to be used in fxci. Great idea!

jwmossmoz:
Once worker-pools.yml and worker-images.yml are removed, this can be merged alongside of https://github.com/mozilla-releng/fxci-config/pull/1042.

jwmossmoz (approved):

PR #922, opened 2026-04-02
Use separate resource group per worker pool in Azure

### LINKED ISSUE #8387: Azure provider hits 800 ARM deployment limit in shared resource group
**Describe the bug**

The `azure2` provider puts all VM deployments into a single resource group (`rg-taskcluster-worker-manager-production`), regardless of worker pool or region. Azure has a hard limit of 800 ARM deployments per resource group. We hit 805 on 2026-03-25 around 14:33 UTC, which broke provisioning for `gecko-t/win11-64-25h2-gpu` across all four configured regions (eastus2, westus, westus2, eastus). Every other pool in the same resource group would have been blocked too.

Two error types appeared:

1. **Quota exceeded** (21+ occurrences): `Creating the deployment would exceed the quota of '800'. The current deployment count is '805'.`
2. **OperationPreempted** (4 occurrences): Side effect of Worker Manager retrying after quota failures. Azure cancelled the stale deployments.

The root cause is that the `azure2` provider creates one ARM deployment per VM, and every deployment goes into the same resource group. ARM deployment records persist after the VM is created and count toward the 800 limit. Once the total across all pools and regions passes 800, nothing new can provision.

**To Reproduce**

1. Go to [Worker Manager errors for gecko-t/win11-64-25h2-gpu](https://firefox-ci-tc.services.mozilla.com/worker-manager/gecko-t%2Fwin11-64-25h2-gpu/errors)
2. Observe 21+ quota errors and 4 OperationPreempted errors from 2026-03-25 14:33-14:45 UTC
3. All errors reference resource group `rg-taskcluster-worker-manager-production` with deployment count 805/800

Affected config:
- Worker pool: `gecko-t/win11-64-25h2-gpu` (all pools in the RG are affected)
- Resource group: `rg-taskcluster-worker-manager-production`
- VM size: `Standard_NV12ads_A10_v5`
- Priority: Spot
- Image: `/subscriptions/108d46d5-fe9b-4850-9a7d-8c914aa6c1f0/resourceGroups/rg-packer-worker-images/providers/Microsoft.Compute/galleries/win11_64_25h2/images/win11_64_25h2/versions/1.0.0`
- Launch configs: `lc-7381ecd4ebc7bead9c7b`, `lc-b1d251af841465d1792d`, `lc-649ffe6adad643106c0a`, `lc-ef32a168786348bcbe06`

**Expected behavior**

Worker Manager should be able to provision VMs without hitting the 800 ARM deployment limit. Old deployment records should be cleaned up after VMs are running, or deployments should be spread across multiple resource groups.

**Taskcluster version**

v97.1.0

**Screenshots**

N/A

**Platform (please complete the following information):**

- OS: N/A (server-side Azure provisioning issue)
- Browser: N/A
- Version: N/A

**Additional context**

References:
- [Worker Manager errors for gecko-t/win11-64-25h2-gpu](https://firefox-ci-tc.services.mozilla.com/worker-manager/gecko-t%2Fwin11-64-25h2-gpu/errors)
- [Azure deployment limit docs](https://aka.ms/800LimitFix)
- [ARM deployment quota exceeded](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/deployment-quota-exceeded)

lotas:
Fixes: https://github.com/taskcluster/taskcluster/issues/8387 Currently every deployment is being applied to the default resource group, which leads to occasional "800 max deployments" limit during spikes of activity.
By using separate resource groups the chances of hitting those limits are significantly lower (unless we request 800+ workers at once in a single pool)

Resource group names are generated as `rg-tc-wp-name` and truncated at 90 chars

W-m is already ensuring those RGs are created if they don't exist

### DIFF
--- a/src/ciadmin/generate/worker_pools.py
+++ b/src/ciadmin/generate/worker_pools.py
@@ -5,6 +5,7 @@
 import copy
 import hashlib
 import json
+import re
 
 import attr
 from tcadmin.resources import WorkerPool
@@ -110,6 +111,15 @@ def _normalize_arm_parameters(parameters):
     }
 
 
+def _sanitize_pool_id_for_rg(pool_id):
+    """Sanitize a pool_id for use in an Azure resource group name.
+
+    Azure RG names: max 90 chars, alphanumeric, hyphens, underscores, periods,
+    parentheses. Pool IDs contain '/' which must be replaced.
+    """
+    return re.sub(r"[^a-zA-Z0-9\-_.]", "-", pool_id)
+
+
 def _build_arm_template_launch_config(
     *,
     pool_id,
@@ -171,11 +181,17 @@ def _build_arm_template_launch_config(
         "parameters": parameters,
     }
 
+    sanitized_pool = _sanitize_pool_id_for_rg(pool_id)
+    arm_resource_group = f"rg-tc-{sanitized_pool}"
+    # Azure RG names max 90 chars
+    arm_resource_group = arm_resource_group[:90]
+
     launch_config = {
         "location": loc,
         "tags": merge(tags),
         "workerConfig": merge(worker_config),
         "armDeployment": arm_deployment,
+        "armDeploymentResourceGroup": arm_resource_group,
         "workerManager": merge(worker_manager_config),
     }
 
--- a/tests/ciadmin/test_generate_worker_pools.py
+++ b/tests/ciadmin/test_generate_worker_pools.py
@@ -262,6 +262,8 @@ def assert_azure_arm(pool):
         "vmSize": {"value": "Standard_F8s_v2"},
     }
 
+    assert launch_config["armDeploymentResourceGroup"] == "rg-tc-provId-my-worker-pool"
+
     assert "hardwareProfile" not in launch_config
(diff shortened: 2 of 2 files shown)


matt-boris (approved):
Thanks Yarik 👍🏻

src/ciadmin/generate/worker_pools.py
```diff
@@ -171,11 +181,17 @@ def _build_arm_template_launch_config(
         "parameters": parameters,
     }
 
+    sanitized_pool = _sanitize_pool_id_for_rg(pool_id)
+    arm_resource_group = f"rg-tc-{sanitized_pool}"
+    # Azure RG names max 90 chars
+    arm_resource_group = arm_resource_group[:90]
```
jcristau, line 187: Is there a risk this makes the names non-unique?  (And would that be a problem?)

lotas: Shouldn't be an issue, all pools are currently short enough to fit, plus even when truncated they are going to produce different names, so deployments would still be spread out across multiple RGs (which is the point)

jcristau (approved):

PR #925, opened 2026-04-09
Bug 2030733 - Implement a new `git-push` tasks_for

ahal:
There are cases where we want to run tasks where Taskcluster Github will not receive events:

1. When pushing a Git notes ref (because Github itself doesn't emit any webhook)
2. When pushing to a user fork of a project (because Taskcluster Github is not installed there)

Both these use cases are needed to help support "Github try pushes". The former because that's how we're going to transfer the try metadata the Decision task needs to know which tasks to schedule. The latter because pushing to forks is how most people work on Github, and it will be a lot nicer for them if they can just continue doing that without needing to push to a dedicated try repo for testing independently of their PRs.

This PR adds a new `build-decision git-push` subcommand, the ability to generate `git-push` hooks that call it, and a `replay-git-push` helper in fxci.

### DIFF
--- a/build-decision/src/build_decision/git_push.py
+++ b/build-decision/src/build_decision/git_push.py
@@ -1,0 +1,45 @@
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at http://mozilla.org/MPL/2.0/.
+
+import json
+import logging
+import os
+
+from .decision import render_tc_yml
+
+logger = logging.getLogger(__name__)
+
+
+def build_decision(*, repository, dry_run):
+    logging.info("Running build-decision task")
+
+    payload = json.loads(os.environ["HOOK_PAYLOAD"])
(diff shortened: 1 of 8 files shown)


build-decision/src/build_decision/git_push.py
```diff
@@ -0,0 +1,55 @@
...
+        if label not in slugids:
+            slugids[label] = slugid.nice()
+        return slugids[label]
+
+    task = render_tc_yml(
+        tc_yml,
+        taskcluster_root_url=os.environ["TASKCLUSTER_ROOT_URL"],
+        tasks_for="git-push",
+        event=event,
+        as_slugid=as_slugid,
```
bhearsum: I'm not seeing anything in the template (either directly in the yml or in the example payloads) that requires this. Is this just here to accommodate tests, or am I missing something?

ahal: Nope you're not missing anything. That was me not catching unnecessary Claude "features". I think I was confusing this with the Github hook templates I've also been working on.

Thanks for catching!

bhearsum (approved):

PR #930, opened 2026-04-15
feat(reviewer-assignment-prod): enable mozilla-firefox/firefox support (bug 2008888)

shtrom:
* feat(reviewer-assignment-production): enable mozilla-firefox/firefox support (bug 2008888)
 * feat(grants): extend shtrom's permissions to all project-engwf
 * reviewer-assignment: use TC indexed images

### DIFF
--- a/grants.d/grants.yml
+++ b/grants.d/grants.yml
@@ -1955,8 +1955,8 @@
 - grant:
     - assume:hook-id:project-engwf/reviewer-assignment-staging
     - hooks:modify-hook:project-engwf/reviewer-assignment-staging
-    - secrets:get:project/engwf/reviewer-assignment/runtime-staging
-    - secrets:set:project/engwf/reviewer-assignment/runtime-staging
+    - secrets:get:project/engwf/*
+    - secrets:set:project/engwf/*
   to:
     - roles:
         - login-identity:mozilla-auth0/ad|Mozilla-LDAP|omehani
--- a/grants.d/grants.yml
+++ b/grants.d/grants.yml
@@ -1953,10 +1953,10 @@
 
 # Bug 2024347: grant shtrom ability to create / modify review assigner hook
 - grant:
-    - assume:hook-id:project-engwf/reviewer-assignment-infra-testing
-    - hooks:modify-hook:project-engwf/reviewer-assignment-infra-testing
-    - secrets:get:project/engwf/reviewer-assignment/runtime-infra-testing
-    - secrets:set:project/engwf/reviewer-assignment/runtime-infra-testing
+    - assume:hook-id:project-engwf/reviewer-assignment-staging
+    - hooks:modify-hook:project-engwf/reviewer-assignment-staging
+    - secrets:get:project/engwf/reviewer-assignment/runtime-staging
+    - secrets:set:project/engwf/reviewer-assignment/runtime-staging
   to:
     - roles:
         - login-identity:mozilla-auth0/ad|Mozilla-LDAP|omehani
--- a/hook-templates/project-engwf/reviewer-assignment-production.yml
+++ b/hook-templates/project-engwf/reviewer-assignment-production.yml
@@ -1,0 +1,54 @@
+---
+created:
+  $fromNow: 0 seconds
+deadline:
+  $fromNow: 2 hours
+expires:
+  $fromNow: 1 month
+extra: {}
+metadata:
+  name: Reviewer assignment
+  description: Assign reviewers for PRs in the mozilla-firefox org
+  owner: omehani@mozilla.com
+  source: 'https://github.com/mozilla-conduit/reviewer-selector'
+payload:
+  capabilities: {}
+  env:
+    TC_SECRET_ID: project/engwf/reviewer-assignment/runtime-prod
+    DIFF_URL:
+      $eval: payload.body.pull_request.diff_url
+    PR_URL:
+      $eval: payload.body.pull_request.html_url
+    ORG_NAME:
+      # e.g., mozilla-firefox
+      $eval: payload.organization
+    REPO_NAME:
+      # e.g., firefox
+      $eval: payload.repository
+    TARGET_BRANCH_NAME:
+      # e.g., autoland
+      $eval: payload.body.pull_request.base.ref
+    REPO_URL:
+      $eval: payload.body.repository.html_url
+
+  features:
+    taskclusterProxy: true
+  image:
+    type: indexed-image
+    namespace: mozilla.v2.reviewer-selector.branch.main.latest.docker-image.reviewer-selector
+    path: public/image.tar.zst
+  maxRunTime: 600
+priority: lowest
+provisionerId: code-review
+retries: 3
+routes:
+  - index.project.engwf.production.reviewer-assignment.latest
+  - index.project.engwf.production.reviewer-assignment.${payload.organization}.${payload.repository}.latest
+  - index.project.engwf.production.reviewer-assignment.${payload.organization}.${payload.repository}.pull-request.${payload.number}
+  - index.project.engwf.production.reviewer-assignment.${payload.organization}.${payload.repository}.revision.${payloa.sha}
+schedulerId: engwf
+scopes:
+  - 'secrets:get:project/engwf/reviewer-assignment/runtime-prod'
+  - 'index:insert-task:project.engwf.production.reviewer-assignment.*'
+tags: {}
+workerType: bot  # XXX create dedicated worker (bug 2024115)
--- a/hook-templates/project-engwf/reviewer-assignment-staging.yml
+++ b/hook-templates/project-engwf/reviewer-assignment-staging.yml
@@ -8,13 +8,13 @@ expires:
 extra: {}
 metadata:
   name: Reviewer assignment
-  description: Assign reviewers for PRs in the mozilla-firefox/infra-testing repo
+  description: Assign reviewers for PRs in the mozilla-firefox staging repos (infra-testing)
   owner: omehani@mozilla.com
   source: 'https://github.com/mozilla-conduit/reviewer-selector'
 payload:
   capabilities: {}
   env:
-    TC_SECRET_ID: project/engwf/reviewer-assignment/runtime-infra-testing
+    TC_SECRET_ID: project/engwf/reviewer-assignment/runtime-staging
     DIFF_URL:
       $eval: payload.body.pull_request.diff_url
     PR_URL:
@@ -33,20 +33,22 @@ payload:
 
   features:
     taskclusterProxy: true
-  # XXX: push to the right place
-  image: docker.io/omehani/moz-reviewer-selector
-  maxRunTime: 7200
+  image:
+    type: indexed-image
+    namespace: mozilla.v2.reviewer-selector.branch.staging.latest.docker-image.reviewer-selector
+    path: public/image.tar.zst
+  maxRunTime: 600
 priority: lowest
 provisionerId: code-review
 retries: 3
 routes:
-  # XXX: need to check this
-  - index.project.engwf.production.reviewer-assignment.latest
+  - index.project.engwf.staging.reviewer-assignment.latest
+  - index.project.engwf.staging.reviewer-assignment.${payload.organization}.${payload.repository}.latest
+  - index.project.engwf.staging.reviewer-assignment.${payload.organization}.${payload.repository}.pull-request.${payload.number}
+  - index.project.engwf.staging.reviewer-assignment.${payload.organization}.${payload.repository}.revision.${payloa.sha}
 schedulerId: engwf
 scopes:
-  - 'notify:email:*'
-  - 'secrets:get:project/engwf/reviewer-assignment/runtime-infra-testing'
-  # XXX: need to check this
-  - 'index:insert-task:project.engwf.production.reviewer-assignment.*'
+  - 'secrets:get:project/engwf/reviewer-assignment/runtime-staging'
+  - 'index:insert-task:project.engwf.staging.reviewer-assignment.*'
 tags: {}
 workerType: bot  # XXX create dedicated worker (bug 2024115)
--- a/hooks.yml
+++ b/hooks.yml
@@ -286,3 +286,110 @@ project-engwf/reviewer-assignment-staging:
       repository:
         type: string
         pattern: "^infra-testing$"
+
+project-engwf/reviewer-assignment-production:
+  name: Reviewer assignment hook (production)
+  description: Automatically assign reviewers to PRs
+  owner: omehani@mozilla.com
+  email_on_error: true
+  scopes:
+    - index:insert-task:project.engwf.production.reviewer-assignment.*
+    - queue:create-task:lowest:code-review/bot  # XXX: will need to change when a dedicated worker is setup (bug 2024115)
+    - queue:route:index.project.engwf.production.reviewer-assignment.*
+    - queue:scheduler-id:engwf
+    - secrets:get:project/engwf/reviewer-assignment/runtime-production
+  template_file: hook-templates/project-engwf/reviewer-assignment-production.yml
+  bindings:
+    - exchange: exchange/taskcluster-github/v1/pull-request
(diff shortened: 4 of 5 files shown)


hook-templates/project-engwf/reviewer-assignment-prod.yml
```diff
@@ -45,7 +45,7 @@ routes:
 schedulerId: engwf
 scopes:
   - 'notify:email:*'
-  - 'secrets:get:project/engwf/reviewer-assignment/runtime-infra-testing'
+  - 'secrets:get:project/engwf/reviewer-assignment/runtime-prod'
   # XXX: need to check this
```
jcristau: Did the other XXXs get resolved?

hooks/project-engwf/reviewer-assignment-prod.yml
jcristau: Consider moving the image to a more official place before enabling for prod?

hooks.yml
```diff
@@ -228,7 +228,7 @@ project-relman/code-review-production:
             type: string
             description: Taskcluster Task Group ID triggering the hook (should be a Try Task Group)
 
-project-engwf/reviewer-assignment-infra-testing:
+project-engwf/reviewer-assignment-prod:
```
jcristau: Would it be useful to keep the testing hook around alongside the prod one?

grants.d/grants.yml
```diff
@@ -1904,10 +1904,10 @@
 
 # Bug 2024347: grant shtrom ability to create / modify review assigner hook
 - grant:
-    - assume:hook-id:project-engwf/reviewer-assignment-infra-testing
-    - hooks:modify-hook:project-engwf/reviewer-assignment-infra-testing
-    - secrets:get:project/engwf/reviewer-assignment/runtime-infra-testing
-    - secrets:set:project/engwf/reviewer-assignment/runtime-infra-testing
+    - assume:hook-id:project-engwf/*
+    - hooks:modify-hook:project-engwf/*
```
jcristau: I think we'd prefer if this repo remained the source of truth for the (prod) hook?

hook-templates/project-engwf/reviewer-assignment-prod.yml
shtrom: Not yet, but I have bugs for them.

hooks/project-engwf/reviewer-assignment-prod.yml
shtrom: Yep, I have a bug for that, too! https://bugzilla.mozilla.org/show_bug.cgi?id=2028701

hooks.yml
shtrom: Not this time, I don't think: infra-testing is supposed to use all the prod infra at the exception of what repo it uses, so this rename should make most sense.

We'll probably need to add a dev one next time we fiddle, though (:

grants.d/grants.yml
shtrom: I'm happy to reduce to `secrets:set:project/engwf/reviewer-assignment/runtime-prod` (:

shtrom: Ah! But I also need to delete the `-infra-testing` secret afterwards, hence the wildcard.

hook-templates/project-engwf/reviewer-assignment-prod.yml
shtrom: Hum, actually, most have. But I'm not sure about how the indexing work, hence this one.

grants.d/grants.yml
shtrom: How about:
```suggestion
    - assume:hook-id:project-engwf/*
    - secrets:get:project/engwf/*
    - secrets:set:project/engwf/*
```

jcristau: I'm fine with the wildcards, just not with the `modify-hook` scope.

shtrom: Ah, once it's in prod. Makes sense.

shtrom: Suggestion above updated 👆🏼.

hooks.yml
shtrom: Actually... Yeah, let's keep it.

shtrom:
@jcristau here's an updated version of the prod hook, with fewer grants and a TC-indexed image (actual namespace TBC), and keeping the infra-testing separate (still using my dockerhub image, until I work out the correct namespace).

WDYT?

shtrom:
> actual namespace TBC
> ...
> until I work out the correct namespace.

I think I managed to get the indices I want: https://firefox-ci-tc.services.mozilla.com/tasks/ICKTa3HUQ8-ew9L4yyxP-g

hook-templates/project-engwf/reviewer-assignment-prod.yml
```diff
@@ -0,0 +1,54 @@
...
+  maxRunTime: 7200
+priority: lowest
+provisionerId: code-review
+retries: 3
+routes:
+  # XXX: need to check this
+  - index.project.engwf.production.reviewer-assignment.latest
+schedulerId: engwf
+scopes:
+  - 'notify:email:*'
```
ahal: Can we make this the specific e-mail address that it's supposed to notify? Not really an issue or anything, but it's easier to determine what e-mail is actually being used.

hooks.yml
```diff
@@ -270,3 +274,102 @@ project-engwf/reviewer-assignment-infra-testing:
...
+    - queue:scheduler-id:engwf
+    - secrets:get:project/engwf/reviewer-assignment/runtime-prod
+  template_file: hook-templates/project-engwf/reviewer-assignment-prod.yml
+  bindings:
+    - exchange: exchange/taskcluster-github/v1/pull-request
+      routing_key_pattern: '#.mozilla-firefox.#.opened'  # e.g., primary.mozilla-firefox.firefox.opened or primary.mozilla-firefox.firefox.synchronize
+    - exchange: exchange/taskcluster-github/v1/pull-request
+      routing_key_pattern: '#.mozilla-firefox.#.reopened'
+    - exchange: exchange/taskcluster-github/v1/pull-request
+      routing_key_pattern: '#.mozilla-firefox.#.synchronize'  # e.g., primary.mozilla-firefox.firefox.opened or primary.mozilla-firefox.firefox.synchronize
```
ahal, line 308: This is a neat way to filter out PR actions you don't care about!

Fyi normally we handle this in the template itself, e.g:
https://github.com/mozilla-releng/fxci-config/blob/2d76ad4e6ea3b34f83d48a4c2a994b90ce76518d/.taskcluster.yml#L113

But this avoids rendering the template at all. It may still be worth having a guard like the above in the template though (just in case)

hook-templates/project-engwf/reviewer-assignment-prod.yml
ahal: Also AIUI, this scope is only necessary if you're calling the notify service via API directly. It shouldn't be needed if you're using those routes.

hooks.yml
```diff
@@ -270,3 +274,102 @@ project-engwf/reviewer-assignment-infra-testing:
...
+
+project-engwf/reviewer-assignment-prod:
+  name: Reviewer assignment hook (testing)
+  description: Automatically assign reviewers to PRs
+  owner: omehani@mozilla.com
+  email_on_error: true
+  scopes:
+    - assume:project:taskcluster:mozilla-github-repository  # XXX: not sure this is needed anymore
+    - index:insert-task:project.engwf.production.reviewer-assignment.*
+    - notify:email:*
```
ahal: You probably don't need this, and you wouldn't want to use *both* the notify API and notify routes together. I'd recommend sticking with the routes, less code to maintain.

ahal (changes requested):
Thanks, this is looking good! Just a few minor suggestions

hooks.yml
ahal: Now that you removed this scope from the templates, you should also remove it here. This is what actually grants the permission to use it.

It's not a sensitive scope or anything, but better to not grant stuff that isn't being used.

ahal (changes requested):
Thanks, this lgtm, but there's one scope from `hooks.yml` that should be removed before landing

hooks.yml
shtrom: > Now that you removed this scope from the templates, you should also remove it here. This is what actually grants the permission to use it.
> 
> It's not a sensitive scope or anything, but better to not grant stuff that isn't being used.

Cool, done.

What about `assume:project:taskcluster:mozilla-github-repository`?

EDIT: I also removed `queue:route:notify.email`, but am now wondering if that's the notify API you mentioned?

grants.d/grants.yml
```diff
@@ -1943,7 +1943,7 @@
     - assume:hook-id:project-engwf/reviewer-assignment-infra-testing
     - hooks:modify-hook:project-engwf/reviewer-assignment-infra-testing
     - secrets:get:project/engwf/reviewer-assignment/runtime-infra-testing
-    - secrets:set:project/engwf/reviewer-assignment/runtime-infra-testing
+    - secrets:set:project/engwf/*
```
jcristau, line 1959: might as well also update the `secrets:get:` scope to match

hook-templates/project-engwf/reviewer-assignment-infra-testing.yml
```diff
@@ -33,20 +33,19 @@ payload:
 
   features:
     taskclusterProxy: true
-  # XXX: push to the right place
-  image: docker.io/omehani/moz-reviewer-selector
+  image:
+    type: indexed-image
+    namespace: mozilla.v2.reviewer-selector.branch.develop
```
jcristau: I'd prefer something like `mozilla.v2.reviewer-selector.branch.develop.latest.docker-image.reviewer-selector`:
- tack on the task's kind and name
- add `latest` so it can also be indexed by revision/commit if needed

hook-templates/project-engwf/reviewer-assignment-production.yml
```diff
@@ -0,0 +1,51 @@
...
+      $eval: payload.body.pull_request.base.ref
+    REPO_URL:
+      $eval: payload.body.repository.html_url
+
+  features:
+    taskclusterProxy: true
+  image:
+    type: indexed-image
+    namespace: mozilla.v2.reviewer-selector.branch.main
+    path: public/image.tar.zst
```
jcristau, line 39: This is a pre-existing problem for the other hooks defined here, but as a heads-up this can break if the referenced indexed task expires (i.e. there hasn't been a commit on reviewer-selector/main for a year, typically).

hook-templates/project-engwf/reviewer-assignment-prod.yml
```diff
@@ -0,0 +1,51 @@
...
+    REPO_URL:
+      $eval: payload.body.repository.html_url
+
+  features:
+    taskclusterProxy: true
+  image:
+    type: indexed-image
+    namespace: mozilla.v2.reviewer-selector.branch.main
+    path: public/image.tar.zst
+  maxRunTime: 7200
```
jcristau: Consider reducing this, 2 hours seems excessive.

hook-templates/project-engwf/reviewer-assignment-production.yml
```diff
@@ -0,0 +1,51 @@
...
+  maxRunTime: 7200
+priority: lowest
+provisionerId: code-review
+retries: 3
+routes:
+  - index.project.engwf.production.reviewer-assignment.latest
```
jcristau, line 45: Should we index by PR number?

hooks.yml
```diff
@@ -179,6 +179,10 @@ lint/pre-commit-{version}:
             type: string
             description: Version of pre-commit to install (e.g. "4.0.0"). Defaults to latest.
 
+#
+# EngWG Reviewer Selection
```
jcristau: `s/WG/WF/`?

hooks.yml
```diff
@@ -272,3 +274,100 @@ project-engwf/reviewer-assignment-infra-testing:
...
         pattern: "^infra-testing$"
+
+
+project-engwf/reviewer-assignment-prod:
+  name: Reviewer assignment hook (testing)
+  description: Automatically assign reviewers to PRs
+  owner: omehani@mozilla.com
+  email_on_error: true
+  scopes:
+    - assume:project:taskcluster:mozilla-github-repository  # XXX: not sure this is needed anymore
```
jcristau: pretty sure it isn't

hook-templates/project-engwf/reviewer-assignment-infra-testing.yml
```diff
@@ -33,20 +33,19 @@ payload:
...
+    path: public/image.tar.zst
   maxRunTime: 7200
 priority: lowest
 provisionerId: code-review
 retries: 3
 routes:
-  # XXX: need to check this
   - index.project.engwf.production.reviewer-assignment.latest
```
jcristau: `production` here seems wrong?

jcristau (changes requested):

grants.d/grants.yml
shtrom: I need to set them, but not read the back, so best keep them hidden for prod (:

hooks.yml
shtrom: Heh, I'm not sure this works... I have this https://github.com/mozilla-releng/fxci-config/blob/e9138c744be7fc35a6d19c9184ca74d8b0d285ea/hooks.yml#L222-L225, but the hook still gets triggered on `action: closed`.

shtrom: 😑 it's because I'm missing one level of nesting

shtrom: ... or not. I'm very confused 😅

grants.d/grants.yml
shtrom: Actually, this may cause issues with the TC UI if trying to get to the page of a pre-existing secret for the purpose of updating it. I'll give `set:...*`

hook-templates/project-engwf/reviewer-assignment-production.yml
ahal: Yeah, I recommended this as a short term solution to get off the ground, but we'll likely want to implement a task on the `reviewer-selector` repo that pushes the image to a registry and references it from there instead.

hook-templates/project-engwf/reviewer-assignment-staging.yml
```diff
@@ -0,0 +1,54 @@
+---
```
ahal, line 1: Do we actually need a whole new template for staging? Without actually looking at the differences, it feels like something that would be easy enough to handle with the JSON-e?

shtrom: (Note that it's just renaming the existing infra-testing template). 

Most of the time, we don't need the templates to be different, no. Even the hook definition is very similar barring a few key parameters.

One point in favour of a separate staging definition would be the ability to alter it ahead of time without affecting prod. (Noting, too, that I retain the permission to change the staging template manually in the TC UI, which is not the case for prod).

hook-templates/project-engwf/reviewer-assignment-production.yml
```diff
@@ -2,19 +2,19 @@
 created:
   $fromNow: 0 seconds
 deadline:
-  $fromNow: 2 hours
+  $fromNow: 10 minutes
```
jcristau: I'd leave the longer deadline.  Means it can still run if there's a delay in getting a worker, or the task fails for an intermittent reason and you need to rerun.

jcristau (approved):

hook-templates/project-engwf/reviewer-assignment-staging.yml
ahal: That makes sense and wfm!

ahal (approved):

ahal:
Looks like there are merge conflicts, please update and you likely want to address the deadline comment to. I can merge when ready.

projects.yml
```diff
@@ -1543,6 +1543,10 @@ reviewer-selector:
   branches:
     - name: main
       level: 3
+    - name: staging
+      level: 1
+    - name: develop
+      level: 1
```
jcristau, line 1549: You'll need to update the project's .taskcluster.yml as well, at least the level and the `short_head_ref == "main"` check.

shtrom: Yep: https://github.com/mozilla-conduit/reviewer-selector/pull/13

PR #934, opened 2026-04-16
refactor: remove unused taskcluster_yml_project feature

ahal:
Afaict this was an attempt at providing a shared `.taskcluster.yml` across repos which over time became unused.

As of today, we can now trigger hooks directly from `.taskcluster.yml` files, so if there's a need to share a file across projects, that's almost certainly going to be the better mechanism to do so.


hneiva (approved):
LGTM

PR #936, opened 2026-04-16
Implement hook variants; Rename hook templates; Refactor hooks.yml

ahal:
This stack is in preparation for using the new Taskcluster Github hooks feature. Basically we need a way to support some form of versioning, and variants will allow us to do that.


src/ciadmin/generate/hooks.py
```diff
@@ -3,27 +3,70 @@
...
 
-        task = await get_ciconfig_file(hook.template_file)
+        with open(hook.template_file) as f:
+            # Replace {word} with the matching attribute, leaving unrecognized
+            # placeholders (e.g. JSON-e's {$eval: ...}) untouched.
+            content = _TEMPLATE_RE.sub(
+                lambda m: hook.attributes.get(m.group(1), m.group(0)),
+                f.read(),
+            )
+            task = yaml.safe_load(content)
```
Eijebong: Not a big fan of this. I guess you don't want to use proper JSON-e here (i.e `${variant}`) because it'd be rendering JSON-e in JSON-e which would require `$$` escaping everything?

src/ciadmin/generate/hooks.py
```diff
@@ -3,27 +3,70 @@
...
+            attributes = hook.attributes.copy()
+            attributes.update(variant)
+
+            hook_name = f"{hook.hook_group_id}/{hook.hook_id}"
+
+            fields = {
+                "template_file": hook.template_file,
+            }
+            for field in fields:
+                resolve_keyed_by(fields, field, hook_name, **attributes)
```
Eijebong, line 56: Couldn't this be:

```py
template_file = evaluate_keyed_by(hook.template_file, hook_name, attributes).format(**attributes)
```

Eijebong: (or drop the .format and do it later to stay consistent with other fields)

src/ciadmin/generate/hooks.py
```diff
@@ -3,27 +3,70 @@
...
+
 
 async def update_resources(resources):
     """
     Manage custom hooks.  This file interprets `hooks.yml` in fxci-config.
     Its behavior is largely documented in the comment in that file.
     """
 
-    hooks = await HookConfig.fetch_all()
+    hooks = list(generate_hook_variants(await HookConfig.fetch_all()))
```
Eijebong: nit: I don't think you need to materialize the list here and you can keep hooks as a generator?

tests/ciadmin/test_generate_hooks.py
```diff
@@ -0,0 +1,148 @@
...
+    [
+        (
+            "template_file",
+            {"by-env": {"production": "tmpl-prod.yml", "testing": "tmpl-test.yml"}},
+            "production",
+            "tmpl-prod.yml",
+        ),
+        (
+            "name",
+            {"by-env": {"production": "Hook (prod)", "testing": "Hook (test)"}},
```
Eijebong: Looks like this doesn't work given the failure? Might want to extend the `by-` resolving to all fields? Given my other comment, I think something like this might work?

```py
hook_group_id = evaluate_keyed_by(hook.hook_group_id, hook_name, attributes)
hook_id = evaluate_keyed_by(hook.hook_id, hook_name, attributes)
name = evaluate_keyed_by(hook.name, hook_name, attributes)
description = evaluate_keyed_by(hook.description, hook_name, attributes)
template_file = evaluate_keyed_by(hook.template_file, hook_name, attributes)
scopes = [evaluate_keyed_by(s, hook_name, attributes).format(**attributes) for s in hook.scopes]

yield attr.evolve(
    hook,
    hook_group_id=hook_group_id.format(**attributes),
    hook_id=hook_id.format(**attributes),
    name=name.format(**attributes),
    description=description.format(**attributes),
    template_file=template_file.format(**attributes),
    scopes=scopes,
    attributes=attributes,
    variants=[{}],
)
```

ahal: Ah my bad, I initially had everything supporting `by-key` but then decided to drop it and change it to only explicit keys, then forgot to update the test

src/ciadmin/generate/hooks.py
ahal: Yeah.. I'm not a fan of it either. But I'm also not a fan of double rendering them with JSON-e, like you mentioned it would require escaping `$` everywhere. I could be convinced double JSON-e is better as it is what `hg-push-template.yml` and friends already use (though I was planning to convert them to this system in a follow-up). That being said those templates are very simple compared to the Decision task template I'm planning to add here, so I think it'll get way more confusing.

Some other options:

1. Use another format like `%(val)s`
2. Continue not pre-rendering these hooks, and define a separate template file for each "variant", which results in a lot of duplication. I don't  think this would be an acceptable long term solution, but it could be an immediate term solution if we wanted to punt the question a little bit further down the road. But it's not like it'll be hard to switch, so I vote not this.

src/ciadmin/generate/hooks.py
ahal: It could, just figured this makes it easier to add new fields that get resolved in the future and it's more consistent with the worker-pools variants implementation. If you feel strongly I don't mind changing it

src/ciadmin/generate/hooks.py
ahal: Another reason to avoid double JSON-e, is that it provides a nice guard rail on making your templates too complex.

E.g, I'm planning to have something like:
```
name: project-releng/pre-commit-{version}
variants:
    - version: v1
    - version: v2
```

If the template starts requiring complex logic in the pre-render, then I feel like it'll start becoming hard to maintain and it would be better to just use a separate template at that point. So not supporting JSON-e there forces us to do that.

Though maybe I'm just worrying for nothing.

ahal: I switched to `string.Template`, which I think is better. It'll now raise on unknown substitutions and continue to pass `${...}` and `{$eval: ...}` through unchanged.

Eijebong (approved):
Thanks for getting rid of the naive regex replace

PR #944, opened 2026-04-22
Bug 2028701 - add grants for mozilla-conduit / proj-engwf trust_domain

shtrom:


grants.d/engwf.yml
```diff
@@ -0,0 +1,17 @@
...
+
+# Grant these for all conduit(proj-engwf) repos on github
+- grant:
+    - notify:email:*
+    - queue:route:garbage.*
+    - queue:route:index.garbage.*
+    - queue:route:notify.email.*
+    - queue:route:index.project.engwf.*
+    - queue:route:index.project.engwf.*
+    - queue:create-task:{priority}:{trust_domain}/*
```
ahal, line 13: This breaks security boundaries, you need to have a level in the queue name. But this scope is already granted as part of the `trust-domain-scopes` feature:
https://github.com/mozilla-releng/fxci-config/blob/main/grants.d/grants.yml#L926

Same for the index, though in this case it's a subset of what you're granting here, so I'd be curious to know what route is actually required.

ahal (changes requested):

ahal:
I think in general I'd recommend you remove the blanket grants to `mozilla-conduit` and instead explicitly grant to the repo you need. The blanket grants were just meant to support getting something up and running quickly. But the intention was always that projects relying on those would then move to an explicit grant soon afterwards.

shtrom:
Closing in favour of https://github.com/mozilla-releng/fxci-config/pull/945

PR #945, opened 2026-04-22
Bug 2028701 - Create explicit roles for 'mozilla-conduit/reviewer-sel…

ahal:
…ector'

We'll need to wait for some version of:
https://github.com/mozilla-conduit/reviewer-selector/pull/5

To land before we can merge this.

### DIFF
--- a/projects.yml
+++ b/projects.yml
@@ -1514,6 +1514,23 @@ crash-ping-ingest:
     taskgraph-cron: true
     trust-domain-scopes: true
 
+# Bug 2028701: Setup `mozilla-conduit/reviewer-selector` with Taskcluster
+reviewer-selector:
+  repo: https://github.com/mozilla-conduit/reviewer-selector
+  repo_type: git
+  trust_domain: mozilla
+  trust_project: reviewer-selector
+  branches:
+    - name: main
+      level: 3
+  features:
+    github-taskgraph: true
+    github-pull-request:
+      policy: public
+    pr-actions: true
+    taskgraph-actions: true
+    trust-domain-scopes: true
+
 # WARNING: These grants apply to entire Github orgs.
 #
 # We should only grant the mozilla trust domain and level 1 context here, as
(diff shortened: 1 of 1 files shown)


projects.yml
```diff
@@ -1514,29 +1514,32 @@ crash-ping-ingest:
...
-#
-# We should only grant the mozilla trust domain and level 1 context here, as
-# these are both already treated as insecure. This context is appropriate to
-# run builds, tests and lints, but wouldn't have the capability to use any
-# release infrastructure or have access to any sensitive scopes.
-
-mozilla:
-  repo: https://github.com/mozilla/*
+# Bug 2028701: Setup `mozilla-conduit/reviewer-selector` with Taskcluster
+mozilla-conduit:
```
shtrom: We should probably use a tighter namespace

```suggestion
mozilla-conduit-reviewer-selector:
```

projects.yml
```diff
@@ -1514,29 +1514,32 @@ crash-ping-ingest:
...
+  repo: https://github.com/mozilla-conduit/reviewer-selector
   repo_type: git
   trust_domain: mozilla
+  trust_project: reviewer-selector
   branches:
     - name: main
-      level: 1
+      level: 3
```
shtrom, line 1525: ... ok 😆 

Is this associated to SCM levels?

shtrom:
> We'll need to wait for some version of:
> [mozilla-conduit/reviewer-selector#5](https://github.com/mozilla-conduit/reviewer-selector/pull/5)

Landed.

projects.yml
ahal: no, but that's where the name comes from. Basically it separates resources (worker pools, secrets, index routes, etc) for untrusted vs trusted contexts.

E.g, a level 1 task from a PR wouldn't be able to run on a level 3 worker. So there would be no chance of a malicious actor poisoning the worker cache and influencing security sensitive tasks

projects.yml
ahal: Oh, good catch. Normally we just do the repo name

bhearsum (approved):

PR #951, opened 2026-04-23
feat: add a hook to run pre-commit

ahal:


ahal:
This was tested successfully here:
https://github.com/mozilla-releng/stage-taskcluster-test

I think we might want to make a custom docker image that has pre-commit pre-installed, but I figure that can be done in a follow-up.

I'll replace pre-commit.ci with this hook in a follow-up PR (so we can actually see it working in the PR itself).

worker-pools.yml
```diff
@@ -4581,3 +4581,19 @@ pools:
...
+      minCapacity: 1
+      maxCapacity: 10
+      regions: *default-gcp-regions
+      image: ubuntu-2404-headless
+      implementation: generic-worker/linux-d2g
+      instance_types:
+        - minCpuPlatform: Intel Cascadelake
+          disks:
+            - *persistent-disk-75gb
+          machine_type: n2-standard-2
```
hneiva: Last I checked spot pricing, c3d-standard-4 is basically ~the same price as n2-standard-2 with double mem+cores.

hneiva (approved):
just a nit about instance type, up to you if you want to switch or not.
LGTM

PR #954, opened 2026-04-28
Add opt-in per-region resource group for Azure ARM deployments

jwmossmoz:
Continues the work from [taskcluster/taskcluster#8386](https://github.com/taskcluster/taskcluster/issues/8386). That issue reported the 800-deployments-per-RG cap being hit on the shared `rg-taskcluster-worker-manager-production` RG; #922 closed it by giving each pool its own deployment RG (`rg-tc-<pool>`). This PR adds an **opt-in** flag that splits per region (`rg-tc-<pool>-<loc>`), so a single pool's RG can't fill to the cap during a burst.

Default behavior is unchanged. The flag is enabled per pool in `worker-pools.yml`:

```yaml
config:
  armDeploymentResourceGroupPerRegion: true
```

## Why opt-in

Azure caps subscriptions at **980 resource groups, hard limit, not adjustable** ([source](https://learn.microsoft.com/azure/azure-resource-manager/management/azure-subscription-service-limits)). Across all the FXCI Azure pool families (gecko-t / gecko-1 / gecko-2 / comm-t / comm-1 / enterprise-t / enterprise-1 / nss-t / nss-1 / mozilla-1 / taskgraph-t) × their suffix matrices × ~14 regions each, a global rollout would add several hundred RGs at once. That's a lot of subscription budget to spend without first validating the change on the pool that actually needs it.

Opt-in lets us:
- enable on `gecko-t/win11-64-25h2` first (where the spike happened)
- watch live behavior for a week or two
- expand to other pools as evidence supports it

## Why this fix exists

`gecko-t/win11-64-25h2` got hit today: 1,373 worker requests in 60 minutes, deployment count up to 907 (past the 800 cap), Azure rejecting new deployments:

> Creating the deployment 'deploy-…' would exceed the quota of '800'. The current deployment count is '907'.

Worker-manager already deletes its own deployment records as the last step of the per-worker reap chain (`services/worker-manager/src/providers/azure/index.js:2137-2152`), and v99.2.0 made the chain faster. But during a spike, deployments arrive instantly while deletes wait on the reaper. The cap trips before the reaper catches up. Spreading deployments across per-region RGs bounds the per-RG arrival rate to a single region's share of pool weight — for a 14-region pool, that's roughly 14% of pool traffic in the busiest region.

## Diff

The RG-naming logic moved into `_arm_deployment_resource_group(pool_id, loc, per_region=False)`. When `per_region=False` (default), behavior matches #922 exactly. When True, the pool segment is truncated as needed; `loc` always survives intact so the split holds even for pathologically long pool IDs.

Parametrized regression test covers default mode (with 90-char truncation), opt-in mode (current shape and long-pool truncation). All 4 azure provider tests pass.

## For review

- No terraform changes — worker-manager auto-creates the RG on first use (per #922).
- No IAM changes — service principal has Contributor at subscription scope.
- No effect on existing pools until the flag is set on them. This PR doesn't enable it anywhere; flipping the flag on `gecko-t/win11-64-25h2` is a separate one-line change.
- Existing per-pool RGs go orphan when a pool is opted in. Their deployment history clears on its own; empty RGs can be deleted later.
- Region-removal cleanup for opted-in pools isn't covered here. Separate follow-up.

Independent of #946 — either can land first.

### DIFF
--- a/src/ciadmin/generate/worker_pools.py
+++ b/src/ciadmin/generate/worker_pools.py
@@ -182,8 +182,6 @@ def _build_arm_template_launch_config(
     }
 
     sanitized_pool = _sanitize_pool_id_for_rg(pool_id)
-    # Per-region-per-pool deployment RG so Azure's 800-deployments-per-RG
-    # quota is bounded by single-region arrival rate, not pool-wide.
     arm_resource_group = f"rg-tc-{sanitized_pool}-{loc}"
     # Azure RG names max 90 chars
     arm_resource_group = arm_resource_group[:90]
--- a/src/ciadmin/generate/worker_pools.py
+++ b/src/ciadmin/generate/worker_pools.py
@@ -182,7 +182,9 @@ def _build_arm_template_launch_config(
     }
 
     sanitized_pool = _sanitize_pool_id_for_rg(pool_id)
-    arm_resource_group = f"rg-tc-{sanitized_pool}"
+    # Per-region-per-pool deployment RG so Azure's 800-deployments-per-RG
+    # quota is bounded by single-region arrival rate, not pool-wide.
+    arm_resource_group = f"rg-tc-{sanitized_pool}-{loc}"
     # Azure RG names max 90 chars
     arm_resource_group = arm_resource_group[:90]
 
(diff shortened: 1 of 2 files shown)


lotas:
Commented in slack, will also duplicate here: 
I missed an opportunity earlier to make this balancing of resource groups on the worker manager side. Idea is that instead of per pool/per region overrides, we tell w-m to use X pools. and it would go in round-robin or one by one to create new deployment.
This way no matter how busy a pool is, it would be spread out across X pools. Kind of like launch config selection currently works

ahal (approved):
Thanks, looks reasonable to me.

PR #966, opened 2026-05-04
ci: enable shared pre-commit hook

ahal:


ahal:
Couple things to note here:

1. The Firefox-CI task failed because this PR is also adding the scopes, so it'll pass on the next PR (I manually applied this on staging)
2. I forgot that this repo has both instances installed.. so this PR will cause a pre-commit task to run twice, one per instance. I think that's ok? There's currently no way to filter hooks by instance or anything.. but maybe we can implement that down the road. Alternatively we could leave the pre-commit task disabled here and continue using pre-commit.ci or Github Actions.

ahal:
Sigh, oops:
https://github.com/mozilla-releng/fxci-config/pull/980

bhearsum (approved):

ahal:
Sigh, adding `queue:get-artifact:public/*` to the hook's role didn't appear to get rid of the TC Github failures trying to fetch the files :(

ahal:
The fix to prevent TC-Github from commenting on commits with scope errors landed on Taskcluster main, but missed the latest release.

I'll hold off landing these PRs until it gets deployed to fxci to keep our e-mail spam to a minimum.

PR #975, opened 2026-05-05
fix: explicitly add anonymous role to hook scopes

ahal:
Taskcluster Github makes an authenticated request to fetch public/github/customCheckRunText.md. Even though this is a public artifact, the request hits a 401 because it's authenticated with a bewit url and so the anonymous role doesn't apply.

It's unclear whether this is a bug or not, but in the meantime, explicitly granting the anonymous role to the hook role should make it work.


bhearsum (approved):
Tangentially related issue where bewit's are included when they don't need to be: https://github.com/taskcluster/taskcluster/issues/6724

PR #988, opened 2026-05-12
Bug 2028701 - grants: allow docker os-group scope to reviewer-selector pull requests

shtrom:
This is necessary to build the docker image based on PRs and pushes.

### DIFF
--- a/grants.d/grants.yml
+++ b/grants.d/grants.yml
@@ -2364,3 +2364,10 @@
     - project:
         alias: mozilla-release
         job: action:merge-automation
+
+- grant:
+    - generic-worker:os-group:<..>/docker
+  to:
+    - project:
+        alias: reviewer-selector
+        job: action:github-pull-request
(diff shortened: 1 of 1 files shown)


grants.d/grants.yml
```diff
@@ -2364,3 +2364,10 @@
     - project:
         alias: mozilla-release
         job: action:merge-automation
+
+- grant:
+    - generic-worker:os-group:{trust_domain}-t/t-linux-*/docker
```
shtrom, line 2369: Currently using `mozilla-t/t-linux-2404-wayland` workers, but this is liable to change to a more suitable worker as we refine the MVP.

grants.d/grants.yml
```diff
@@ -2364,3 +2364,10 @@
...
+  to:
+    - project:
+        alias: reviewer-selector
+        job: action:github-pull-request
```
shtrom, line 2373: Would this also cover pushes to `main`?

jcristau (changes requested):
granting the docker os group scope is not accaptable.  To run things in docker use docker-worker task payloads.
taskgraph has a docker-image kind and image_builder image that deal with this.

PR #1006, opened 2026-05-20
fix: include response body when Github is unhappy about GITHUB_TOKEN

JohanLorenzo:
## Context

I occasionally run `ci-admin`  these days. When I do, I sometimes forget to set the `GITHUB_TOKEN` which I quickly remember thanks to the existing warning. However, I also have to reissue a new token which means the SAML config of my token gets reset. With the current behavior, I have no way of knowing what is wrong with my `GITHUB_TOKEN` except by remembering what I did last time. 

With this PR, I hope that I'll be able to remember faster next time 🙂

## Verification

Without `GITHUB_TOKEN` (rate limit hit):
```
WARNING: GITHUB_TOKEN is not present in the environment; you may run into rate limits querying for GitHub branches
Got error when querying /repos/mozilla-releng/fxci-config/branches: 403 rate limit exceeded: {"message":"API rate limit exceeded for 1.2.3.4. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

With `GITHUB_TOKEN` not SSO-authorized for the `mozilla` org:
```
Got error when querying /repos/taskcluster/taskgraph/branches: 403 Forbidden: {"message":"Resource protected by organization SAML enforcement. You must grant your Personal Access token access to this organization.","documentation_url":"https://docs.github.com/articles/authenticating-to-a-github-organization-with-saml-single-sign-on/","status":"403"}
```

### DIFF
--- a/tests/ciadmin/test_generate_branches.py
+++ b/tests/ciadmin/test_generate_branches.py
@@ -1,0 +1,76 @@
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at http://mozilla.org/MPL/2.0/.
+
+import json
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import aiohttp
+import pytest
+
+from ciadmin.generate import branches as branches_module
+
+
+def make_mock_client(response):
+    client = AsyncMock()
+    client.request = AsyncMock(return_value=response)
+    return client
+
+
+def make_403_response(body_text):
+    response = MagicMock()
+    response.ok = False
+    response.status = 403
+    response.reason = "Forbidden"
+    response.text = AsyncMock(return_value=body_text)
+    response.raise_for_status.side_effect = aiohttp.ClientResponseError(
+        MagicMock(), MagicMock(), status=403
+    )
+    return response
+
+
+@pytest.mark.asyncio
+async def test_get_403_rate_limit_exceeded(capsys):
+    branches_module._cache.clear()
+    body = json.dumps(
+        {
+            "message": "API rate limit exceeded for 1.2.3.4. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
+            "documentation_url": "https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting",
+        }
+    )
+    response = make_403_response(body)
+    client = make_mock_client(response)
+
+    with patch(
+        "ciadmin.generate.branches.github.get_client", AsyncMock(return_value=client)
+    ):
+        with pytest.raises(aiohttp.ClientResponseError):
+            await branches_module.get("mozilla-releng/fxci-config")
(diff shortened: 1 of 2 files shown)


tests/ciadmin/test_generate_branches.py
```diff
@@ -0,0 +1,80 @@
...
+    response.raise_for_status.side_effect = aiohttp.ClientResponseError(
+        MagicMock(), MagicMock(), status=403
+    )
+    return response
+
+
+@pytest.fixture(autouse=True)
+def clear_cache():
+    branches_module._cache.clear()
+    branches_module._lock.clear()
```
bhearsum: Clearing out the lock is completely unnecessary...there's no harm in them already existing in the dictionary, and the locks release at the end of `get`. And with that removed there's no reason to wrap the cache clearing in a fixture IMO.

tests/ciadmin/test_generate_branches.py
```diff
@@ -0,0 +1,80 @@
...
+    response.status = 403
+    response.reason = "Forbidden"
+    response.text = AsyncMock(return_value=body_text)
+    response.raise_for_status.side_effect = aiohttp.ClientResponseError(
+        MagicMock(), MagicMock(), status=403
+    )
+    return response
+
+
+@pytest.fixture(autouse=True)
```
bhearsum: This is somewhat a preference thing, so not blocking, but IMO it's much more clear to be explicit about fixture usage rather than `autouse`. With the latter, you need additional context outside of a specific test case to understand the full picture.

JohanLorenzo: True. I removed it.

tests/ciadmin/test_generate_branches.py
JohanLorenzo: Sounds good. No more lock then 👌

tests/ciadmin/test_generate_branches.py
```diff
@@ -0,0 +1,79 @@
...
+    response.text = AsyncMock(return_value=body_text)
+    response.raise_for_status.side_effect = aiohttp.ClientResponseError(
+        MagicMock(), MagicMock(), status=403
+    )
+    return response
+
+
+@pytest.fixture
+def clear_cache():
+    branches_module._cache.clear()
```
bhearsum: I think I was less than clear in my earlier comment; I was trying to say that there's no reason to wrap this clearing in a fixture -- it can be done directly in the tests instead. (This way adds unnecessary indirection.)

Feel free to merge with or without changing this, it's not really a big deal.

bhearsum (approved):

tests/ciadmin/test_generate_branches.py
JohanLorenzo: Oh, my bad. I removed the fixture entirely then. It wasn't a big thing to change 🙂

PR #1014, opened 2026-05-25
fix: don't allow untrusted pull requests to be granted scopes to secrets

bhearsum:
https://github.com/mozilla-mobile/mozilla-vpn-client/pull/11315 highlighted the fact that untrusted pull requests are still granted scopes to L1 secrets. There may be certain cases where this is OK, but it certainly shouldn't be the default. This change adjusts what I believe to be a bug in the `trust-domain-scope` grants, and changes the secrets granted there to exclude untrusted pull requests.

The diff for this looks sane, although it's quite difficult to read because it involves the creation of new roles for so many projects.

### DIFF
--- a/grants.d/grants.yml
+++ b/grants.d/grants.yml
@@ -913,10 +913,6 @@
     - secrets:get:project/taskcluster/gecko/hgfingerprint
     - secrets:get:project/taskcluster/gecko/hgmointernal
 
-    # Allow fetching secrets appropriate to this level
-    - secrets:get:project/{trust_domain}/level-{level}/*
-    - secrets:get:project/{trust_domain}/level-any/*
-
     # Allow a sensible scheduler-id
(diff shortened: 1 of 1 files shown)


grants.d/grants.yml
```diff
@@ -933,6 +929,16 @@
...
 
+- grant:
+    # Allow fetching secrets appropriate to this level
+    - secrets:get:project/{trust_domain}/level-{level}/*
+    - secrets:get:project/{trust_domain}/level-any/*
+  to:
+    - project:
+        feature: trust-domain-scopes
+        # Secrets are granted everywhere _except_ untrusted pull requests
+        job: [action:*, branch:*, cron:*, pr-action:*, pull-request:trusted, release:*]
```
hneiva, line 940: would be cool to have grant/deny. Kinda like IAM roles

hneiva (approved):
LGTM

PR #1015, opened 2026-05-26
revert: trust-domain-scope grants for secrets (#1014)

bhearsum:
This was done in haste, and didn't account for earlier conversations where it was decided we were OK with this (L1 secrets do not need to be protected the way L3 ones do; the correct solution if something of value is stored in them is to remove it).


bhearsum:
We decided to keep the changes in #1014. If there are cases where we want specific secrets to be accessible to untrusted pull requests we will grant those specifically and selectively rather than rely on blanket access to L1 secrets.

PR #1019, opened 2026-05-27
Bug 2008888 - projects: set reviewer-selector main builds in level 1

shtrom:


shtrom:
Abandoned in favour of https://github.com/mozilla-conduit/reviewer-selector/pull/10/.

If the build is for prod-artifacts, we do want to build them in a level 3 context.

PR #1024, opened 2026-05-28
fix(reviewer-assignment): set correct per-PR routes (bug 2022122)

shtrom:


shtrom:
Now the hook doesn't fail (at this point) https://firefox-ci-tc.services.mozilla.com/tasks/TsmfVxzGSnWpc6MYOQ1GWw (the next failure is due to https://github.com/mozilla-conduit/reviewer-selector/pull/13)

jcristau (approved):

PR #1027, opened 2026-05-29
feature(grants): add mozillians-group:engworkflow grants to project-engwf (bug 2042692)

shtrom:
Using this Mozillians group: https://people.mozilla.org/a/engworkflow.


grants.d/grants.yml
```diff
@@ -1949,15 +1949,17 @@
...
 
-# Bug 2024347: grant shtrom ability to create / modify review assigner hook
+# Bug 2042692: grant engwf permissions to their project namespaces
 - grant:
+    # Bug 2024347: ability to create / modify review assigner hook
     - assume:hook-id:project-engwf/reviewer-assignment-staging
     - hooks:modify-hook:project-engwf/reviewer-assignment-staging
+
     - secrets:get:project/engwf/*
     - secrets:set:project/engwf/*
```
jcristau: The engworkflow group already has these scopes via https://github.com/mozilla-releng/fxci-config/blob/9ff5148dec93753790bb638fa3dd8f879780614d/grants.d/grants.yml#L808-L814, so either merge the two blocks or remove the secrets grants from here?

shtrom: Ah, interesting. Is the `engworkflow` group the same as the `engworkflow` mozillians group?

I wasn't in the latter until last week, which is likely why I needed explicit grants to the same thing up to now.

Anyway, I merged it all.

grants.d/grants.yml
```diff
@@ -809,9 +809,16 @@
...
+
+    # Bug 2024347: ability to create / modify review assigner hook
+    - assume:hook-id:project-engwf/reviewer-assignment-staging
+    - hooks:modify-hook:project-engwf/reviewer-assignment-staging
+
   to:
     - groups:
         - engworkflow
+    - roles:
+        - mozillians-group:engworkflow
```
shtrom, line 821: May not be needed if it the same as the `engworkflow` group above.

If so, delete at will.

```suggestion
```

grants.d/grants.yml
jcristau: Yes.  https://github.com/mozilla-releng/fxci-config/blob/751e5594a52493cfe7f2b2e136de8a7560a0f9d1/grants.d/grants.yml#L2064-L2068 defines `groups: [engworkflow]` to be the same as `roles: [mozillians-group:engworkflow]` (we should remove this IMO, and use the role throughout)

shtrom: Which one should we remove? That definition you linked to?

And do you want me to take care of it here?

jcristau: Yes please.

grants.d/grants.yml
```diff
@@ -809,9 +809,14 @@
...
     - secrets:get:project/engwf/*
     - secrets:set:project/engwf/*
+
+    # Bug 2024347: ability to create / modify review assigner hook
+    - assume:hook-id:project-engwf/reviewer-assignment-staging
+    - hooks:modify-hook:project-engwf/reviewer-assignment-staging
+
   to:
-    - groups:
-        - engworkflow
```
shtrom, line 816: This was the only occurence.

grants.d/grants.yml
shtrom: Done!

ahal (approved):

shtrom:
@ahal could you merge this PR? I don't have the permission (:

PR #1029, opened 2026-06-03
Add enterprise-thunderbird

lissyx:


ahal (approved):

lissyx:
(do not merge yet there are discussions on how to do things ongoing)

ahal:
We also need all branches that this references to exist and a `.taskcluster.yml` to exist on them before we can merge.

lissyx:
> We also need all branches that this references to exist and a `.taskcluster.yml` to exist on them before we can merge.

it's verified to be the case

ahal:
Looks like it's still failing with:
>  aiohttp.client_exceptions.ClientResponseError: 404, message='Not Found', url='https://api.github.com/repos/mozilla/enterprise-thunderbird-try/contents/.taskcluster.yml?ref=enterprise-main'

lissyx:
Ah because there is no `enterprise-thunderbird-try` yet: do we want one?

projects.yml
```diff
@@ -664,6 +664,53 @@ enterprise-firefox-try:
...
+      level: 1
+  features:
+    gecko-roles: true
+    git-push: true
+    github-taskgraph: true
+    github-pull-request:
+      policy: collaborators
+    pr-actions: true
+    scriptworker: true
+    taskgraph-actions: true
```
jcristau: We shouldn't have action hooks for the "try" repo I think?  Only prod repos/branches should get hooks created (that can then be used by the "try" repo if .tc.yml matches).

jcristau (dismissed):

projects.yml
lissyx: I think I copy/pasted existing ones and just changed names, so we would need to fix on enterprise-firefox-try?

lissyx: @jcristau should I remove `pr-actions`, `taskgraph-actions` or both?

ahal:
> Ah because there is no `enterprise-thunderbird-try` yet: do we want one?

I think it could be a separate PR if it isn't necessary quite yet, but if you know for sure it'll be needed soon creating it is fine.. We could also just have them re-use `enterprise-try` I don't think we need a whole new try repo per trust domain

projects.yml
ahal: You can remove both

lissyx:
> > Ah because there is no `enterprise-thunderbird-try` yet: do we want one?
> 
> I think it could be a separate PR if it isn't necessary quite yet, but if you know for sure it'll be needed soon creating it is fine.. We could also just have them re-use `enterprise-try` I don't think we need a whole new try repo per trust domain

removed for now

lissyx:
The current failure is likely expected. From looking at https://github.com/mozilla-releng/fxci-config/blob/7cc6ebe04c71450264c914686a941c32baa5a88f/src/ciadmin/check/check_pull_request_policies.py#L56 and then https://github.com/mozilla-releng/fxci-config/blob/7cc6ebe04c71450264c914686a941c32baa5a88f/src/ciadmin/check/check_pull_request_policies.py#L15-L22

It just downloads the .taskcluster.yml which for now does not have:
```
policy:
    pullRequests: collaborators
```

Is it required to be here right now? It will be there with my changes after.

ahal:
> Is it required to be here right now? It will be there with my changes after.

Yes, if this task is failing it blocks deploys.

ahal (changes requested):
Checks can't be failing or they block deploys

lissyx:
I've done the change, does it needs to be on all branches?

ahal (approved):

PR #1036, opened 2026-06-11
fix: add mozilla + pre-commit grants to the taskcluster org

ahal:
This works around an issue where various Github events in the Taskgraph repo didn't have scopes to trigger the pre-commit hook.

We technically don't need to grant `mozilla` scopes to the whole org to do this, but it doesn't hurt to make bootstrapping new repos here a bit more turnkey.

### DIFF
--- a/projects.yml
+++ b/projects.yml
@@ -742,7 +742,6 @@ taskgraph:
     github-pull-request:
       policy: public_restricted
     pr-actions: true
-    pre-commit: true
     taskgraph-actions: true
     trust-domain-scopes: true
     treeherder-reporting: true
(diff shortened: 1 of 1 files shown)


projects.yml
```diff
@@ -1592,3 +1591,18 @@ mozilla-releng:
     pre-commit: true
     taskgraph-actions: true
     trust-domain-scopes: true
+
+taskcluster:
+  repo: https://github.com/taskcluster/*
+  repo_type: git
+  trust_domain: mozilla
```
bhearsum, line 1598: It's a bit weird to me that this org is in the `mozilla` trust domain while taskgraph is in the `taskgraph` trust domain. Can we move it to the `mozilla` trust domain?

ahal: Maybe it could be in the `releng` trust domain? I agree it doesn't need a dedicated trust domain, but we should treat `mozilla` as stuff that doesn't need strong security.

The confusing part might be that `mozilla` trust domain isn't meant to reference the mozilla org in Github. E.g the `mozilla-releng` org also has the "mozilla" trust domain, but we have tons of repos there that use something different

bhearsum: Oh, the confusing part to me was that we'd have repos that have scopes from two different trust domain with this change. I didn't realize this was already the case with `mozilla-releng` repos. So...meh.

bhearsum (approved):
(I'm still not sure I'm high on having repos with scopes from multiple trust domains, but that ship has sailed for now, so there's no reason to hold back on this for that reason.)

projects.yml
ahal: Oh, I see!

Personally I think we should be sharing pools across trust domains more, at least for things that don't need security. Maybe we can invent some standard to accomplish this that's separate from a "trust domain" though.

PR #1039, opened 2026-06-12
feat(workflows): Add firefox .tc.yml watch workflow

hneiva:
Triggers a deploy workflow whenever the .taskcluster.yml file in the firefox repo changes.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -6,9 +6,24 @@ on:  # yamllint disable-line rule:truthy
   pull_request:
     branches: ["main"]
   workflow_dispatch: {}
+  workflow_call:
+    inputs:
+      trigger_reason:
+        description: "Human-readable note describing what triggered this deploy."
+        type: string
+        required: false
+        default: ""
 
 permissions:
   contents: read
+
+# Serialize every deploy into a single group so only one ever runs at a time.
+# A trigger arriving mid-deploy queues one pending run rather than starting a
+# concurrent apply. cancel-in-progress is false so an in-flight apply is never
+# interrupted.
+concurrency:
+  group: deploy
+  cancel-in-progress: false
 
 jobs:
   check:
@@ -39,7 +54,7 @@ jobs:
         run: uv run tc-admin check --environment ${{matrix.environment}}
 
   apply:
-    if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'
+    if: github.event_name == 'push' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule'
     runs-on: ubuntu-latest
     needs: check
     environment: apply-${{matrix.environment}}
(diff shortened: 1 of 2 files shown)


bhearsum:
I haven't actually reviewed the patch, but this seems like an excellent way to improve things re: the need for manual hook rebuilding! I suggest that at the very least, `autoland` is covered by this as well (it's probably more important than `main` seeing as its where `.taskcluster.yml` changes first land.

And for cross linking purposes, https://bugzilla.mozilla.org/show_bug.cgi?id=1917274#c5 has a plan to solve this robustly for the entire cluster (but that shouldn't detract from taking this easy win).

hneiva:
- switched to autoland (thx for pointing it out 😃)
- added a note about this being a temporary fix and linked to the bug

.github/workflows/deploy.yml
```diff
@@ -6,10 +6,25 @@ on:  # yamllint disable-line rule:truthy
...
+        default: ""
 
 permissions:
   contents: read
 
+# Serialize every deploy into a single group so only one ever runs at a time.
+# A trigger arriving mid-deploy queues one pending run rather than starting a
+# concurrent apply. cancel-in-progress is false so an in-flight apply is never
+# interrupted.
+concurrency:
```
bhearsum, line 24: Really, this is unrelated to what you're doing here, but it's good to have. I think it will fix https://bugzilla.mozilla.org/show_bug.cgi?id=1929281?

.github/workflows/firefox-tc-watch.yml
```diff
@@ -0,0 +1,82 @@
...
+#
+# TEMPORARY: this is a stopgap -- it only monitors the firefox repo's
+# `.taskcluster.yml`, not the equivalent configs across all other Taskcluster
+# projects/repos. A permanent, general solution is tracked in
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1917274.
+#
+# Why autoland (not main): a `.taskcluster.yml` change can break Taskcluster
+# hooks, so the matching fxci-config deploy must land promptly. Changes hit
+# autoland continuously but only merge to main ~twice a day, so watching main
+# could leave hooks broken for hours. We accept that an autoland change may be
```
bhearsum: What purpose does this final sentence serve? (I suggest removing it.)

.github/workflows/firefox-tc-watch.yml
```diff
@@ -0,0 +1,82 @@
...
+# backed out (a later re-deploy then reconverges) in exchange for fast reaction.
+#
+# How it works:
+#   - On a schedule, fetch the SHA of the latest commit touching `.taskcluster.yml`.
+#   - Use actions/cache keyed on that SHA as a "have we already seen this?" marker:
+#       * cache miss  -> new (or first-seen) commit -> run `deploy`
+#       * cache hit   -> already handled            -> do nothing
```
bhearsum: IMO this caching should be removed. Imagine this case:
* `.taskcluster.yml` change lands on autoland
* This runs, caches, and causes the hooks to be rebuilt
* The `.taskcluster.yml` change is broken for a reason that doesn't affect the sha (perhaps it also included changes to taskgraph, for example).
* This runs, caches, and causes the hooks to be rebuilt
* The `.taskcluster.yml` change is fixed in a way that doesn't affect the sha and re-lands on autoland
* Hooks are not rebuilt

Rebuilding is cheap, easy, safe, and (as cited) idempotent -- let's just run it whenever we see a change.

bhearsum (changes requested):
I very much look forward to having this! I think it's probably a good idea to cover other repos and branches as well; I can't imagine we'd be bumping up against rate limits if we did so.

Whether here or in a follow-up (eg: after this is verified to work) I suggest also covering:
* beta, release, esr115, esr140 in the Firefox repo
* `enterprise-{main,release,beta}` in https://github.com/mozilla/enterprise-firefox
* Maybe `main` and other branches in https://github.com/mozilla-mobile/firefox-ios?
* `main` in https://github.com/mozilla-mobile/mozilla-vpn-client

(Most of the other github repos we build hooks for are either ours, owned by someone close to us, or so infrequently used it's probably not worth bothering.)

It would be lovely if we could cover Thunderbird as well, but seeing as they are still on hg, it's probably not viable, and may end up getting served 406s and breaking this action.

.github/workflows/deploy.yml
hneiva: Yeah, this is mostly in case we are deploying via merge, and the cron wants to update things at the same time. It's a super corner case, but easy enough to cover.

.github/workflows/firefox-tc-watch.yml
hneiva: That's not really how the cache works. It's caching the last seen commit hash of the .tc.yml file.
It would rebuild hooks in all 3 steps you mentioned.

bhearsum: So there's only one hash cached at any given time?

In any case, what benefit are we getting from the caching?

hneiva: We only deploy if there's a change to .tc.yml on the "watched" repos

bhearsum: Why do we need to cache hashes to do that? We're filtering on changes that touch `.taskcluster.yml` already, no?

hneiva: we look at the last commit hash that touched the file, if it changes in any way, it's a green light to deploy. The cache is there as a memory between each run.

bhearsum: Oh sorry, this is on me. For some reason I kept assuming this was firing when files changed...but it's on cron.

bhearsum (approved):

PR #1043, opened 2026-06-16
feat: switch reps pull request policy to public

bhearsum:
AFAICT this repository doesn't run any tasks that require secrets, so we can enable public pull requests.

(The template that this tool uses to create other configurations does, but that doesn't matter here...)

### DIFF
--- a/projects.yml
+++ b/projects.yml
@@ -772,7 +772,7 @@ reps:
   features:
     github-taskgraph: true
     github-pull-request:
-      policy: public_restricted
+      policy: public
     taskgraph-actions: true
     trust-domain-scopes: true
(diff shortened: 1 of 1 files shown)


ahal (approved):

ahal:
Is this to fix all the failures from renovate there I guess?

bhearsum:
> Is this to fix all the failures from renovate there I guess?

Yeah, that's what prompted it, but it would potentially hit by PRs from others (although quite unlikely....).

bhearsum:
https://github.com/mozilla-releng/reps/pull/108 for the reps side of things

PR #1045, opened 2026-06-17
reviewer-assignment: use release for prod and main for staging repos (bug 2046275)

shtrom:


shtrom:
> Here's a diff when I ran this locally which seems to show a couple of issues:
> [...]
> 
>     1. substitutions don't seem to work in `bindings`
> 
>     2. `pattern` is getting removed from `organization` and `repository` (maybe that's intentional?)

Ah, great, thanks! How did you run that diff?

1. is a problem, will fix
2. is intentional (it is not functional at the moment, due to https://github.com/taskcluster/taskcluster/issues/8867)

shtrom:
> Here's a diff when I ran this locally

Side question: is this something that could be added to run as a check on PRs, similar to what we have on Terraform changes in infra repos?

bhearsum:
> Ah, great, thanks! How did you run that diff?

`uv run ci-admin diff --environment firefoxci` will do it, except you need a GITHUB_TOKEN and taskcluster credentials with `auth:list-clients` and `hooks:list-hooks:*` scopes in the environment. (I don't know if you have the latter.)

> Side question: is this something that could be added to run as a check on PRs, similar to what we have on Terraform changes in infra repos?

Funny you mention that, this motivated me to take a stab at it in https://github.com/mozilla-releng/fxci-config/pull/1103. We'll see if it pans out.

shtrom:
Yep, I don't have the TC permissions; hoping it DOES pan out (:

shtrom:
Oh, no!

I suspect `bindings` are not updated by the variants logic: 

https://github.com/mozilla-releng/fxci-config/blob/fe34b252cfa41e8f8ad9f131ce3bf834af3fa927/src/ciadmin/generate/hooks.py#L71-L82


🤔 unless it's in the `attributes`... no, doesn't look like it https://github.com/mozilla-releng/fxci-config/blob/16e3d8ac8f06c623778b309e4a3c9139783c25da/src/ciadmin/generate/ciconfig/hooks.py#L22-L23

shtrom:
Ok, I now have #1105 to apply `variants` to `bindings`, which I merged back into this PR to see if that fixes point 2.

We'll have to merge #1105 first.

src/ciadmin/generate/hooks.py
shtrom, line 1: From #1105.

tests/ciadmin/test_generate_hooks.py
shtrom, line 1: From #1105.

jcristau:
> Yep, I don't have the TC permissions; hoping it DOES pan out (:

You should have enough to run the `diff` and `check` sections of
https://docs.mozilla-releng.net/en/latest/how-to/taskcluster/ci_admin.html#run.
And you can pass `--resources hooks` to speed it up.

shtrom:
> > Yep, I don't have the TC permissions; hoping it DOES pan out (:
> 
> You should have enough to run the `diff` and `check` sections of https://docs.mozilla-releng.net/en/latest/how-to/taskcluster/ci_admin.html#run. And you can pass `--resources hooks` to speed it up.

Ah, yep, that worked! Thanks.

Ok, with #1105 merged, the diff now looks like it should.

@bhearsum r?

hooks.yml
```diff
@@ -220,6 +227,7 @@ project-engwf/reviewer-assignment-staging:
         description: |
           The raw body of github event (for version 1)
         required:
+          - action
           - number
           - pull_request
```
bhearsum, line 232: I guess this ends up getting passed for prod hooks as an empty string or some such, and goes unused?

shtrom: 🤔 It was a mismatch between both, and I preferred to err on the side of inclusivity.

However, I think this is a noop, and I will soon no longer use most of those fields, so I'll have to cull it down.

Looks like https://github.com/taskcluster/taskcluster/issues/8867 got merged, too, so it may create many layers of headaches 😅

PR #1042, opened 2026-06-16
feat: migrate fuzzing worker images to fxci

jwmossmoz:
## Summary

- add fuzzing worker image records using TCEng GCP, AWS, and Azure workflow outputs
- restore the fuzzing worker-pool definitions from fxci-config PR #899 before the worker-pool/image revert
- add the matching fuzzing grants so the new `proj-fuzzing` pools are usable
- add the extra Azure GPU regions needed by the fuzzing Windows GPU pools

## Image sources

- GCP generic-worker-ubuntu-24-04: worker-images run 27632282848, `projects/taskcluster-imaging/global/images/generic-worker-ubuntu-24-04-qioehfdraiiishpuvgvl`
- GCP generic-worker-ubuntu-24-04-arm64: worker-images run 27632284378, `projects/taskcluster-imaging/global/images/generic-worker-ubuntu-24-04-arm64-mqscbutmsbawtytlroxk`
- AWS docker-worker legacy: copied from community account `885316786408` into Taskcluster account `692406183521`
  - `us-east-1`: `ami-0a6e926238859761c`
  - `us-east-2`: `ami-00f6c9d7fec4fa42b`
  - `us-west-1`: `ami-03e4f8db63254ce7e`
  - `us-west-2`: `ami-0b5dd0bbb670ec80e`
- AWS generic-worker-ubuntu-24-04 staging: copied from relops-prod account `961225894672` into Taskcluster account `692406183521`
  - `us-east-1`: `ami-02619e55246806e8d`
- Azure generic-worker-win2022: worker-images run 27708116372, Taskcluster `90.0.1`, in `rg-packer-worker-images`
  - `central-us`: `imageset-pytrrvjottfuqbemodpn-centralus-fuzzing`
  - `east-us`: `imageset-qqkmmgddstxfhvbsdprj-eastus-fuzzing`
- Azure generic-worker-win2022-gpu: worker-images run 27708116314, Taskcluster `90.0.1`, in `rg-packer-worker-images`
  - `east-us`: `imageset-fcxybmrwdcsmxjijwamv-eastus-fuzzing`
  - `east-us-2`: `imageset-ofghwxvlubvnimhurgqa-eastus2-fuzzing`
  - `south-central-us`: `imageset-hhgwofqoyyhwaxoucnli-southcentralus-fuzzing`
  - `west-us-2`: `imageset-owexteqnyghtbsfeuppb-westus2-fuzzing`

## Related infrastructure work

[...]

PR #1048, opened 2026-06-18
fix: use ARM template for fuzzing Azure pools

jwmossmoz:
## Summary
- add the standard `taskcluster-arm-template` template spec to the new Azure fuzzing Windows pools
- move `capacityPerInstance` into `worker-manager-config`, which is the shape consumed by the generator

This follows up on #1042. The deploy apply failed because `proj-fuzzing/ci-windows` generated a direct Azure launch config with `storageProfile.imageReference` but no `storageProfile.osDisk`, so worker-manager rejected it against the Azure provider schema. The other new Azure fuzzing pools had the same shape and would fail next.

## Verification
- generated the four affected pool configs directly and asserted every launch config uses `armDeployment`, not direct `storageProfile`
- `uv run pytest tests/ciadmin/test_generate_worker_pools.py`
- `GITHUB_TOKEN=$(gh auth token) uv run tc-admin check --environment firefoxci`

### DIFF
--- a/worker-pools.yml
+++ b/worker-pools.yml
@@ -3579,6 +3579,8 @@ pools:
       worker-purpose: proj-fuzzing
       locations: [central-us, east-us]
       maxCapacity: 10
+      armDeployment:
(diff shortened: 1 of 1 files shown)


aerickson (approved):
lgtm

PR #1049, opened 2026-06-18
fix: use windows generic-worker config for fuzzing pools

jwmossmoz:
## Summary
- switch the Azure fuzzing Windows pools to `generic-worker/windows`
- avoid emitting `runTasksAsCurrentUser`, which the migrated fuzzing images reject at startup

Follows the merged fuzzing work in #899, #1042, and #1048. The smoke tasks now get past pool creation/ARM provisioning but fail when generic-worker loads config.

## Verification
- targeted generation check: affected pools no longer include `runTasksAsCurrentUser` and still use ARM launch configs
- `uv run pytest tests/ciadmin/test_generate_worker_pools.py`
- `GITHUB_TOKEN=$(gh auth token) uv run tc-admin check --environment firefoxci`

### DIFF
--- a/worker-pools.yml
+++ b/worker-pools.yml
@@ -3583,7 +3583,7 @@ pools:
     provider_id: azure2
     config:
       image: fuzzing-gw-win2022
-      implementation: generic-worker/worker-runner-windows
+      implementation: generic-worker/windows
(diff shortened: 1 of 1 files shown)


aerickson (approved):
lgtm

PR #1059, opened 2026-06-25
Grant Thunderbird Lando scopes for l10n bump and release promotion (bug 2050355)

coreycb:


grants.d/grants.yml
```diff
@@ -2366,6 +2366,9 @@
     - project:
         alias: [autoland, mozilla-beta, mozilla-release]
         job: cron:l10n-bumper
+    - project:
+        alias: [comm-central, comm-beta, comm-release, comm-esr140]
```
bhearsum: Not 153 as well? (Or maybe that's coming later, after that repo is more set-up/live?)

bhearsum (approved):

grants.d/grants.yml
coreycb: I think we  could add that now. Let me do that. Thanks for the quick review!

PR #1062, opened 2026-06-29
Migrate proj-taskcluster from community-tc

petemoore:
Migrate proj-taskcluster entities from https://community-tc.services.mozilla.com/ to https://firefox-ci-tc.services.mozilla.com/.

### DIFF
--- a/clients.yml
+++ b/clients.yml
@@ -1349,10 +1349,27 @@ project/relsre/devicepool:
   scopes:
     - queue:claimed-count:*
     - queue:pending-count:*
+
 project/taskcluster/smoketest:
-  description: Run smoketests on new Taskcluster deployments
+  description: For running smoke tests against firefox-ci services (typically when deploying updated serivces)
   scopes:
     - assume:project:taskcluster:smoketests
+project/taskcluster/generic-worker/ci-macos:
+  description: Client for all workers in static workerpool proj-taskcluster/gw-ci-macos.
+  scopes:
+    - assume:worker-pool:proj-taskcluster/gw-ci-macos
+    - assume:worker-id:proj-taskcluster/*
+project/taskcluster/generic-worker/taskcluster-ci:
+  description: |
+    A client used by generic-worker CI tests.
+  scopes:
+    - assume:project:taskcluster:generic-worker-tester
+project/taskcluster/testing/client-libraries:
+  description: |
+    This client is used to test the client libraries in Taskcluster CI.
+  scopes:
+    - "object:upload:taskcluster:taskcluster/test/*"
+    - "object:download:taskcluster/test/*"
 project/taskcluster/audit-reports:
   description: Used for auditing worker versions across pools.
   scopes:
@@ -1360,6 +1377,7 @@ project/taskcluster/audit-reports:
     - queue:scheduler-id:smoketest
     - queue:seal-task-group:smoketest/*
     - queue:get-artifact:public/*
+
 project/wpt/wptsync:
   description: Created in Bug 1596469 - Secret handed to :jgraham
   scopes:
--- a/externally-managed.yml
+++ b/externally-managed.yml
@@ -19,3 +19,9 @@ fuzzing:
   - "Hook=project-fuzzing/.*"
   - "Role=hook-id:project-fuzzing/.*"
   - "Secret=project/fuzzing/.*"
+
+taskcluster:
+  # smoketests create these resources, and also cleans them up, so leave
+  # them alone here...
+  - "Client=project/taskcluster/smoketest/.*"
+  - "Role=project:taskcluster:smoketest:.*"
--- a/grants.d/taskcluster.yml
+++ b/grants.d/taskcluster.yml
@@ -1,0 +1,211 @@
+# Taskcluster project grants
+
+- grant:
+    - notify:manage-denylist
+    # This is the taskcluster channel
+    - queue:route:notify.matrix-room.!whDRjjSmICCgrhFHsQ:mozilla.org.*
+  to:
+    - roles:
+        - project-admin:taskcluster
+
+- grant:
+    - queue:create-task:highest:proj-taskcluster/ci
+    # For replacing the docker-worker pool (proj-taskcluster/ci) for d2g work
+    - queue:create-task:highest:proj-taskcluster/gw-ubuntu-24-04
+    # The account and secret for the Azure testing storage account.
+    # This is secret but ok for use by PRs.
+    - secrets:get:project/taskcluster/testing/azure
+    # This is the taskcluster channel
+    - queue:route:notify.matrix-room.!whDRjjSmICCgrhFHsQ:mozilla.org.*
+  to:
+    - roles:
+        - repo:github.com/taskcluster/*
+
+- grant:
+    - queue:create-task:highest:proj-taskcluster/ci
+  to:
+    - roles:
+        - repo:github.com/mozilla/hawk:*
+
+- grant:
+    - secrets:get:project/taskcluster/testing/client-libraries
+    - secrets:get:project/taskcluster/testing/taskcluster-*
+    - docker-worker:cache:taskcluster-*
+  to:
+    - roles:
+        - repo:github.com/taskcluster/taskcluster:*
+
+- grant:
+    - queue:create-task:highest:proj-taskcluster/release
+    - secrets:get:project/taskcluster/release
+  to:
+    - roles:
+        - repo:github.com/taskcluster/taskcluster:tag:v*
+
+- grant:
+    - queue:create-task:highest:proj-taskcluster/release
+    - secrets:get:project/taskcluster/staging-release
+  to:
+    - roles:
+    # both pre- and post-bug-1635455
+        - repo:github.com/taskcluster/taskcluster:branch:staging-release/*
+        - repo:github.com/taskcluster/staging-releases:*
+
+- grant:
+    # pushes to main and releases can notify on failure
+    - queue:route:notify.email.taskcluster-internal@mozilla.com.*
+    - queue:route:notify.email.taskcluster-notifications@mozilla.com.*
+    - queue:route:notify.irc-channel.#taskcluster-bots.*
+  to:
+    - roles:
+        - repo:github.com/taskcluster/taskcluster:branch:main
+        - repo:github.com/taskcluster/taskcluster:tag:v*
+
+- grant:
+    - queue:create-task:highest:proj-taskcluster/ci
+  to:
+    - roles:
+        - repo:github.com/json-e/json-e:*
+
+- grant:
+    # pushes to json-e main can read secrets for deploying the site
+    - secrets:get:project/taskcluster/json-e-deploy
+    # ..and notify on failure
+    - queue:route:notify.email.taskcluster-notifications@mozilla.com.*
+  to:
+    - roles:
+        - repo:github.com/json-e/json-e:branch:main
+
+- grant:
+    - assume:project:taskcluster:worker-test-scopes
+    - assume:worker-id:docker-worker/docker-worker
+    - assume:worker-id:random-local-worker/docker-worker
+    - docker-worker:cache:docker-worker-garbage-*
+    - docker-worker:capability:device:hostSharedMemory
+    - docker-worker:capability:device:hostSharedMemory:null-provisioner/*
+    - docker-worker:capability:device:loopbackAudio
+    - docker-worker:capability:device:loopbackAudio:null-provisioner/*
+    - docker-worker:capability:device:loopbackVideo
+    - docker-worker:capability:device:loopbackVideo:null-provisioner/*
+    - docker-worker:capability:disableSeccomp
+    - docker-worker:capability:disableSeccomp:null-provisioner/*
+    - docker-worker:capability:privileged
+    - docker-worker:capability:privileged:null-provisioner/*
+    - docker-worker:image:localhost:*
+    - purge-cache:null-provisioner/*
+    - queue:cancel-task
+    - queue:cancel-task-group:taskcluster-level-1/*
+    - queue:cancel-task:docker-worker-tests/*
+    - queue:claim-task
+    - queue:claim-task:null-provisioner/*
+    - queue:claim-work:null-provisioner/*
+    - queue:create-artifact:*
+    - queue:create-task:lowest:null-provisioner/*
+    - queue:create-task:lowest:proj-taskcluster/ci
+    - queue:get-artifact:private/docker-worker-tests/*
+    - queue:rerun-task:taskcluster-level-1/*
+    - queue:resolve-task
+    - queue:route:statuses
+    - queue:scheduler-id:docker-worker-tests
+    - queue:scheduler-id:taskcluster-github
+    - queue:seal-task-group:taskcluster-level-1/*
+    - queue:worker-id:docker-worker/docker-worker
+    - queue:worker-id:random-local-worker/docker-worker
+    - queue:worker-id:random-local-worker/dummy-worker-*
+    - secrets:get:project/taskcluster/taskcluster-worker/stateless-dns
+    - secrets:get:project/taskcluster/testing/docker-worker/ci-creds
+    - secrets:get:project/taskcluster/testing/docker-worker/pulse-creds
+  to:
+    - roles:
+        - repo:github.com/taskcluster/taskcluster:*
+        - project:taskcluster:docker-worker-tester
+
+- grant:
+    - auth:create-client:project/taskcluster/smoketest/*
+    - auth:create-role:project:taskcluster:smoketest:*
+    - auth:delete-client:project/taskcluster/smoketest/*
+    - auth:delete-role:project:taskcluster:smoketest:*
+    - auth:reset-access-token:project/taskcluster/smoketest/*
+    - auth:update-client:project/taskcluster/smoketest/*
+    - auth:update-role:project:taskcluster:smoketest:*
+    - project:taskcluster:smoketest:*
+    - purge-cache:built-in/succeed:smoketest-cache
+    - queue:create-task:highest:built-in/*
+    - queue:create-task:highest:built-in/fail
+    - queue:create-task:highest:built-in/succeed
+    - queue:route:index.project.taskcluster.smoketest.*
+    - queue:scheduler-id:smoketest
+    - secrets:get:project/taskcluster/smoketest/*
+    - secrets:set:project/taskcluster/smoketest/*
+  to:
+    - roles:
+        - project:taskcluster:smoketests
+
+- grant:
+    - assume:project:taskcluster:smoketests
+  to:
+    - roles:
+        - github-team:taskcluster/smoketesters
+
+- grant:
+    - assume:worker-id:test-worker-group/test-worker-id
+    - assume:worker-pool:test-provisioner/*
+    - assume:worker-type:test-provisioner/*
+    - auth:create-client:project/taskcluster:generic-worker-tester/TestReclaimCancelledTask
+    - auth:create-client:project/taskcluster:generic-worker-tester/TestResolveResolvedTask
+    - auth:sentry:generic-worker-tests
+    - docker-worker:cache:d2g-test
+    - docker-worker:capability:privileged:test-provisioner/test-*
+    - docker-worker:capability:device:hostSharedMemory:test-provisioner/test-*
+    - docker-worker:capability:device:kvm:test-provisioner/test-*
+    - docker-worker:capability:device:loopbackAudio
+    - docker-worker:capability:device:loopbackAudio:test-provisioner/test-*
+    - docker-worker:capability:device:loopbackVideo
+    - docker-worker:capability:device:loopbackVideo:test-provisioner/test-*
+    - docker-worker:feature:allowPtrace
+    - generic-worker:cache:apple-cache
+    - generic-worker:cache:banana-cache
+    - generic-worker:cache:devtools-app
+    - generic-worker:cache:test-modifications
+    - generic-worker:cache:unknown-issuer-app-cache
+    - generic-worker:os-group:test-provisioner/*
+    - generic-worker:run-as-administrator:test-provisioner/*
+    - generic-worker:run-task-as-current-user:test-provisioner/*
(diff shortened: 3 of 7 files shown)


grants.d/taskcluster.yml
```diff
@@ -0,0 +1,211 @@
+# Taskcluster project grants
+
+- grant:
+    - notify:manage-denylist
```
jcristau: this doesn't seem appropriate here?

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,211 @@
...
+    # This is the taskcluster channel
+    - queue:route:notify.matrix-room.!whDRjjSmICCgrhFHsQ:mozilla.org.*
```
jcristau: similar, why is this granted to a non-repo role?

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,211 @@
...
+  to:
+    - roles:
+        - project-admin:taskcluster
+
+- grant:
+    - queue:create-task:highest:proj-taskcluster/ci
```
jcristau, line 4: I would prefer if we added projects to projects.yml, and grants were generated from that, instead of directly here.

externally-managed.yml
```diff
@@ -19,3 +19,9 @@ fuzzing:
   - "Hook=project-fuzzing/.*"
   - "Role=hook-id:project-fuzzing/.*"
   - "Secret=project/fuzzing/.*"
+
+taskcluster:
+  # smoketests create these resources, and also cleans them up, so leave
+  # them alone here...
+  - "Client=project/taskcluster/smoketest/.*"
+  - "Role=project:taskcluster:smoketest:.*"
```
jcristau, line 27: I don't think we should have that here, we already have smoketests set up in grants.d/grants.yml and in clients.yml.

jcristau (changes requested):
Didn't do a thorough review, but noting a few things that stood out to me.

grants.d/taskcluster.yml
petemoore: I understand the concern, but note, this is a limitation of the notify service, i.e. orthogonal to the migration. Unfortunately the Notify service doesn't support more granular controls, just a global scope for managing the (global) deny list. A client/role either has the scope or not. Currently in community, this scope is granted to roles `github-team:taskcluster/core` and `project-admin:taskcluster`. Keeping this scope here preserves behaviour when project-taskcluster migrates from community to fxci.

grants.d/taskcluster.yml
petemoore: Is the question why `queue:route:notify.matrix-room.!whDRjjSmICCgrhFHsQ:mozilla.org.*` is granted to `project-admin:taskcluster` rather than to `repo:github.com/taskcluster/*`?

grants.d/taskcluster.yml
jcristau: Preserving the behaviour is not the goal here in my opinion.  It might or might not be appropriate for the taskcluster team to be able to manage the notify denylist on fxci, but that is separate from letting taskcluster's CI run here, and should not be part of the same PR.

grants.d/taskcluster.yml
jcristau: Something like that yes (I'd have concerns about `repo:github.com/taskcluster/*` literally, but, details).

grants.d/taskcluster.yml
petemoore: Can you explain in a bit more detail? In projects.yml I don't see any grants, only trust domains. Do you mean creating a new project trust domain?

externally-managed.yml
petemoore: The intention is not to rearchitect projects during migration, but to lift them as-is. These are the clients and the roles that the existing smoke tests use, so changing this also means migrating tests in taskcluster. If we want to change this, I would propose we do this as a post-migration step.

grants.d/taskcluster.yml
jcristau: Grants are added automatically when you add a project in projects.yml, so yes, my suggestion is to add projects there, and only add extra grants if necessary.  That might not map 1:1 to the naming scheme in community in terms of worker pool names etc, but I don't think that should be a showstopper.

grants.d/taskcluster.yml
petemoore: For posterity, Julien and I met up, and agreed indeed this is not appropriate. Historically the taskcluster team had specific rights which are unrelated to the taskcluster project which essentially governs the taskcluster CI, rather than operational responsibilities that the same taskcluster staff have over taskcluster deployments. I've removed both these grants now.

externally-managed.yml
petemoore: Apologies, you are right! They are already set up, removing here...

grants.d/taskcluster.yml
petemoore: Thanks!

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,176 @@
...
+  to:
+    - roles:
+        - repo:github.com/taskcluster/taskcluster:branch:main
+        - repo:github.com/taskcluster/taskcluster:tag:v*
+
+- grant:
+    - queue:create-task:highest:proj-taskcluster/ci
+  to:
+    - roles:
+        - repo:github.com/json-e/json-e:*
```
ahal, line 60: We'll need the `firefoxci-taskcluster` app installed in this org and added to this repo in case you haven't done it already.

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,176 @@
...
+    - docker-worker:capability:disableSeccomp
+    - docker-worker:capability:disableSeccomp:null-provisioner/*
+    - docker-worker:capability:privileged
+    - docker-worker:capability:privileged:null-provisioner/*
+    - docker-worker:image:localhost:*
+    - purge-cache:null-provisioner/*
+    - queue:cancel-task
+    - queue:cancel-task-group:taskcluster-level-1/*
+    - queue:cancel-task:docker-worker-tests/*
+    - queue:claim-task
```
ahal: What can one do with this scope? Does this give `taskcluster/taskcluster` the ability to claim any task?

Same question for the other scopes like `cancel-task` and `resolve-task`

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,176 @@
...
+    - queue:claim-task:null-provisioner/*
+    - queue:claim-work:null-provisioner/*
+    - queue:create-artifact:*
```
ahal: This is also suspect. Can you namespace what's allowed to be created?

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,176 @@
...
+    - docker-worker:cache:docker-worker-garbage-*
+    - docker-worker:capability:device:hostSharedMemory
+    - docker-worker:capability:device:hostSharedMemory:null-provisioner/*
+    - docker-worker:capability:device:loopbackAudio
+    - docker-worker:capability:device:loopbackAudio:null-provisioner/*
+    - docker-worker:capability:device:loopbackVideo
+    - docker-worker:capability:device:loopbackVideo:null-provisioner/*
+    - docker-worker:capability:disableSeccomp
+    - docker-worker:capability:disableSeccomp:null-provisioner/*
+    - docker-worker:capability:privileged
```
ahal: Can these capabilities be namespaced?

clients.yml
```diff
@@ -1349,17 +1349,35 @@ project/relsre/devicepool:
   scopes:
     - queue:claimed-count:*
     - queue:pending-count:*
+
 project/taskcluster/smoketest:
-  description: Run smoketests on new Taskcluster deployments
+  description: For running smoke tests against firefox-ci services (typically when deploying updated serivces)
```
ahal: I liked the old description better. The `staging` deployment also uses these clients for instance.

worker-images.yml
```diff
@@ -219,7 +219,7 @@ fuzzing-gw-win2022-gpu:
     south-central-us: imageset-wklenlbxaoeghdemjdyk-southcentralus
     west-us-2: imageset-cnnahgrkshgsftrapiac-westus2
 
-fuzzing-gw-win2022:
+gw-win2022:
```
ahal, line 222: I'd like a relops stamp for these

worker-pools.yml
```diff
@@ -3731,3 +3731,279 @@ pools:
...
+            idleTimeoutSecs: 3600
+            shutdownMachineOnInternalError: false
+            runAfterUserCreation: C:\Windows\System32\calc.exe
+      vmSizes:
+        - vmSize: Standard_F8s_v2
+          launchConfig:
+            hardwareProfile:
+              vmSize: Standard_F8s_v2
+      worker-manager-config:
+        capacityPerInstance: 1
```
ahal, line 4009: I think it would be very simple to use our variants system here to eliminate duplication.

Given pools were defined differently in community, these aren't copied over 1 for 1 anyway. I'm sure Claude can figure it out trivially ;)

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,176 @@
...
+    - queue:get-artifact:SampleArtifacts/b/c/d.jpg
+    - queue:resolve-task
+    - queue:scheduler-id:test-scheduler
+  to:
+    - roles:
+        - project:taskcluster:generic-worker-tester
+
+- grant:
+    - assume:project:taskcluster:generic-worker-tester
+    - queue:create-task:highest:proj-taskcluster/*
```
Eijebong, line 151: This allows any PR to create jobs on the release pool.

grants.d/taskcluster.yml
petemoore: I think we should add it after the PR lands. Are you happy with that ordering?

worker-images.yml
petemoore: CC @jwmoss

worker-pools.yml
petemoore: Done

grants.d/taskcluster.yml
petemoore: Fixed

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,240 @@
+# Taskcluster project grants
+#
+# Kept in step with the `taskcluster` project in community-tc-config
+# (config/projects/taskcluster.yml). Deliberate differences: no
+# project-admin grants, no personal-fork roles, and no grants for pools
+# that don't exist here (e.g. proj-taskcluster/gw-ci-macos).
```
ahal: This comment is going to get stale fast, I'd put it in the extended commit message rather than in-repo.

worker-images.yml
jwmossmoz: @petemoore I'm fine with this.

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,240 @@
...
+- grant:
+    # The account and secret for the Azure testing storage account.
+    - secrets:get:project/taskcluster/testing/azure
+    - secrets:get:project/taskcluster/testing/client-libraries
+    - secrets:get:project/taskcluster/testing/taskcluster-*
+  to:
+    - roles:
+        - repo:github.com/taskcluster/taskcluster:pull-request
+        - repo:github.com/taskcluster/taskcluster:branch:*
+        - repo:github.com/taskcluster/taskcluster:tag:*
```
ahal: nit: since this is the same set of roles, you could merge these two grant blocks together

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,240 @@
...
+    - project:
+        alias: taskcluster
+
+- grant:
+    - queue:create-task:highest:proj-taskcluster/ci
+  to:
+    - roles:
+        - repo:github.com/mozilla/hawk:*
+
+# Trusted pull requests, branches, and tags use level 3. Untrusted pull
```
ahal: Having pull requests be L3 isn't good. I guess it's how things have been up until now so I can be convinced to let it slide to help get this landed.

But it should be something we fix very soon afterwards and I'd like us to agree on who will be on the hook for that.

src/ciadmin/check/check_grants.py
```diff
@@ -169,12 +171,25 @@ def is_level_1(role):
...
         return bool(level_1.search(role))
 
+    # The taskcluster/taskcluster CI mirrors community-tc: trusted pull requests
+    # (from collaborators, or untrusted PRs a collaborator explicitly runs via
+    # comment) share the project's own `taskcluster-level-3` namespace with
+    # pushes and tags, while untrusted pull requests are isolated in
+    # `taskcluster-level-1`. These are scheduler-id / task-group namespaces,
+    # not fxci trust levels: they grant no access to level-3 worker pools.
+    taskcluster_trusted_pr = "repo:github.com/taskcluster/taskcluster:pull-request"
+    taskcluster_level_3_ns = re.compile(r":taskcluster-level-3(?:/|$)")
```
ahal: I don't want to normalize this by making it seem like this mode is supported or working ok.

Instead please implement a XFAIL mechanism. Basically define a dict of scopes by role `xfail_scopes` where if it matches we log a warning rather than fail the test.

This serves as a reminder for us to fix them. Bonus points if the test fails when there aren't any matches (e.g a strict xfail).

ahal (changes requested):
Thanks, getting closer! I think the only code change I want to see before approving is fixing the test to make it clear that L3 PRs are not ok.

grants.d/taskcluster.yml
petemoore: Nice spot! Done.

grants.d/taskcluster.yml
petemoore: Agree, this should probably be gated on https://github.com/mozilla-releng/releng-rfcs/pull/57 landing. Once the strategy has been agreed, I am happy to pick this up. I can create an issue for this if you like.

src/ciadmin/check/check_grants.py
petemoore: Done!

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,176 @@
...
+    # For replacing the docker-worker pool (proj-taskcluster/ci) for d2g work
+    - queue:create-task:highest:proj-taskcluster/gw-ubuntu-24-04
+    # This is the taskcluster channel
+    - queue:route:notify.matrix-room.!whDRjjSmICCgrhFHsQ:mozilla.org.*
+  to:
+    - project:
+        alias: taskcluster
+
+# Trusted pull requests, branches, and tags use level 3. Untrusted pull
+# requests use a separate level-1 namespace below. Credentials used by
```
jcristau: Can we make this follow the existing patterns from fxci?  I don't think it's appropriate to have collaborator pull requests share context with production tasks, and be labelled as "level 3".

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,176 @@
...
+    - docker-worker:cache:taskcluster-*
+    # The account and secret for the Azure testing storage account.
+    - secrets:get:project/taskcluster/testing/azure
+    - secrets:get:project/taskcluster/testing/client-libraries
+    - secrets:get:project/taskcluster/testing/taskcluster-*
+  to:
+    - roles:
+        - repo:github.com/taskcluster/taskcluster:pull-request
+        - repo:github.com/taskcluster/taskcluster:branch:*
+        - repo:github.com/taskcluster/taskcluster:tag:*
```
jcristau: I think we should prefer granting scopes to projects rather than roles, e.g.
```
to:
  - projects:
      alias: [taskcluster]
      job: [branch:*, tag:*, pull-request]
````
(the existing `taskcluster/*` entry in projects.yml needs to be revisited)

This applies to a lot of the grants in this file.

**Review by jcristau — CHANGES_REQUESTED** [review]

**Review by ahal — APPROVED** [review]
Thanks for making those changes! Looks much better

File: src/ciadmin/generate/worker_pools.py [hunk 19]
```diff
@@ -280,7 +280,9 @@ def get_aws_provider_config(
                 {"availability-zone": availability_zone},
             )
 
-            for instance_type in sorted(instance_types):
+            for instance_type in sorted(
+                instance_types, key=lambda x: x["instanceType"]
+            ):
```
jcristau: What is this about?

worker-pools.yml
```diff
@@ -169,6 +169,27 @@ worker-defaults:
...
+      # mirrors the imageset's workerConfig in community-tc-config
+      # config/imagesets.yml: the signing key lives at
+      # /etc/generic-worker/ed25519_key (not the monopacker path above) and d2g
+      # is enabled, since Taskcluster's decision task is a docker-worker payload.
+      generic-worker/linux-tceng:
+        genericWorker:
+          config:
+            d2gConfig:
+              enableD2G: true
+            ed25519SigningKeyLocation: /etc/generic-worker/ed25519_key
```
jcristau: Presumably this is a throw-away key, not trusted for anything sensitive?

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,162 @@
+# Taskcluster project grants (github.com/taskcluster/*, json-e/json-e)
+
+# Any repo in the taskcluster GitHub org may use the shared CI pools and
+# notify the team's Matrix room.
+- grant:
+    - queue:create-task:highest:proj-taskcluster/ci
+    - queue:create-task:highest:proj-taskcluster/gw-ubuntu-24-04
+    - queue:route:notify.matrix-room.!whDRjjSmICCgrhFHsQ:mozilla.org.*
```
jcristau: nit: ideally we wouldn't let random PRs notify the channel.  Not a blocker, we can always drop it if/when it becomes a problem.

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,162 @@
...
+    - queue:cancel-task:{trust_domain}-level-{level}/*
+    - queue:rerun-task:{trust_domain}-level-{level}/*
+    - docker-worker:cache:{trust_domain}-level-{level}-*
+    - generic-worker:cache:{trust_domain}-level-{level}-*
+    - queue:route:index.{trust_domain}.cache.level-{level}.*
+    - index:insert-task:{trust_domain}.cache.level-{level}.*
+  to:
+    - project:
+        alias: [taskcluster, taskcluster-staging-releases]
+        job: [tag:*]
```
jcristau: Nice.  We might want to extend the generic grants later to support tag pushes.

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,162 @@
...
+    - docker-worker:capability:device:loopbackAudio
+    - docker-worker:capability:device:loopbackAudio:test-provisioner/test-*
+    - docker-worker:capability:device:loopbackVideo
+    - docker-worker:capability:device:loopbackVideo:test-provisioner/test-*
+    - docker-worker:feature:allowPtrace
+    - generic-worker:cache:apple-cache
+    - generic-worker:cache:banana-cache
+    - generic-worker:cache:devtools-app
+    - generic-worker:cache:test-modifications
+    - generic-worker:cache:unknown-issuer-app-cache
```
jcristau: I don't suppose these can be `test-` or `tc-test-` prefixed e.g.?
Shouldn't matter since the worker pools themselves are namespaced, but it'd match the existing pattern if easy enough to change.

grants.d/taskcluster.yml
```diff
@@ -0,0 +1,162 @@
...
+    - generic-worker:os-group:test-provisioner/*
+    - generic-worker:run-as-administrator:test-provisioner/*
+    - generic-worker:run-task-as-current-user:test-provisioner/*
+    - generic-worker:loopback-audio:test-provisioner/test-*
+    - generic-worker:loopback-video:test-provisioner/test-*
+    - index:find-task:garbage.generic-worker-tests.*
+    - index:insert-task:garbage.generic-worker-tests.*
+    - queue:cancel-task:test-scheduler/*
+    - queue:create-artifact:public/*
```
jcristau: This one looks odd, is `public` the taskId here?

projects.yml
```diff
@@ -1669,7 +1669,7 @@ mozilla-releng:
     taskgraph-actions: true
     trust-domain-scopes: true
 
-taskcluster:
+taskcluster-org:
```
jcristau: Just a note that renaming this causes some scope changes, so if anything relies on `mozilla.v2.taskcluster-pr.*
` index routes in fxci today they'll have to update. (I don't think it'll be a problem, but just in case...)

src/ciadmin/generate/worker_pools.py
petemoore: This is a fix for when AWS pools have more than one instanceType (by chance they currently have only one).

worker-pools.yml
petemoore: Exactly, throw-away unused key, since generic-worker requires one.

grants.d/taskcluster.yml
petemoore: Indeed, and nothing is currently using it, but it enables the flow if needed. Agree we can drop it if it becomes a problem. 👍

grants.d/taskcluster.yml
petemoore: Nice idea

grants.d/taskcluster.yml
petemoore: It is a little bit of work, but not huge. I'm currently working on a PR in taskcluster and community-tc-config for this.

grants.d/taskcluster.yml
petemoore: Good eye! I'll remove it. This is legacy stuff.

projects.yml
petemoore: Agreed, this is safe to change as nothing is using `mozilla.v2.taskcluster-pr` nor `mozilla.v2.taskcluster`. Having `taskcluster-org` for the org and `taskcluster` for the repo hopefully avoids confusion/overloading.

grants.d/taskcluster.yml
petemoore: Created https://github.com/taskcluster/taskcluster/pull/9143 and https://github.com/taskcluster/community-tc-config/pull/1010.

jcristau (approved):
Thanks!

jwmossmoz (approved):

PR #1064, opened 2026-06-30
Updates lando scopes for Thunderbird to tbird lando-repos

stephenking314:


bhearsum (changes requested):
I'm not really sure what the intent is here...but the `alias` blocks need to refer to top-level keys in `projects.yml`.

PR #1065, opened 2026-06-30
Update Thunderbird lando repos

stephenking314:


bhearsum (approved):
I'll assume these names are correct; in the future it would be helpful to point at whatever would help verify that.

PR #1077, opened 2026-07-01
fix(workflows): only serialize the `apply` job

jcristau:
Checks can and should run in parallel, otherwise they end up getting canceled on some PRs.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -17,14 +17,6 @@ on:  # yamllint disable-line rule:truthy
 permissions:
   contents: read
 
-# Serialize every deploy into a single group so only one ever runs at a time.
-# A trigger arriving mid-deploy queues one pending run rather than starting a
-# concurrent apply. cancel-in-progress is false so an in-flight apply is never
-# interrupted.
-concurrency:
-  group: deploy
-  cancel-in-progress: false
-
 jobs:
   check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -58,6 +50,14 @@ jobs:
     runs-on: ubuntu-latest
     needs: check
     environment: apply-${{matrix.environment}}
+    # Serialize only the actual deploy into a single group so only one apply
+    # ever runs at a time. A trigger arriving mid-deploy queues one pending
+    # run rather than starting a concurrent apply. cancel-in-progress is
+    # false so an in-flight apply is never interrupted. Checks (including on
+    # PRs) are unaffected and always run immediately.
+    concurrency:
+      group: deploy-apply
```
Eijebong: Probably want to key that by environment? I don't know how github behaves if you schedule 2 jobs from a matrix with the same group name. Bit scared that it would schedule both and cancel the first one since it doesn't seem to check any kind of provenance on jobs

```suggestion
      group: deploy-apply-${{ matrix.environment }}
```

jcristau: In theory yes, but the apply job has a single environment (firefoxci) anyway.

Eijebong: Yeah but I'd expect to be able to add something to the matrix without having to worry about weird bugs and the fix isn't exactly costly

jcristau: Fairy nuff :)

Eijebong (approved):
Thank you

PR #1082, opened 2026-07-03
fix: tighten up staging landoscript scopes for firefox

bhearsum:
We're going to be adding similar scopes for Thunderbird, and don't want to cross the streams between the staging Thunderbird and Firefox repositories.

### DIFF
--- a/grants.d/grants.yml
+++ b/grants.d/grants.yml
@@ -2332,16 +2332,21 @@
         - mozillians-group:windows-laf-token-managers
 
 ### landoscript scopes
-# on try, we're fairly liberal with these scopes for ease of testing, and
-# allow all actions to be run through any means.
+# on try and in PRs, we're fairly liberal with these scopes for ease of testing,
+# and allow all actions to be run through any means.
 - grant:
     - project:releng:lando:action:*
-    # there are staging versions of all production repos; allow try access
-    # to all of them
-    - project:releng:lando:repo:staging-*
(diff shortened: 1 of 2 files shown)


grants.d/grants.yml
```diff
@@ -2338,10 +2338,17 @@
...
     # to all of them
-    - project:releng:lando:repo:staging-*
+    - project:releng:lando:repo:{lando_repo}
   to:
     - project:
         alias: try
+
+- grant:
+    - project:releng:lando:action:*
+    # there are staging versions of all production repos; allow try access
```
jcristau: The comment says try but the grant is for firefox PRs?

projects.yml
```diff
@@ -462,7 +462,7 @@ cypress:
       - nightly-desktop-win64-aarch64
 try:
   repo: https://hg.mozilla.org/try
-  lando_repo: ff-test-dev
+  lando_repo: staging-firefox-*
```
jcristau, line 468: This seems worth an explanation in the commit message.

jcristau (changes requested):

jcristau (approved):
I think what I was missing was the fact that try's lando_repo wasn't actually used anywhere.

bhearsum:
> I think what I was missing was the fact that try's lando_repo wasn't actually used anywhere.

Ah, yeah. I was thrown by that at first as well (which is part of the reason I wanted to update the grant to use it).

PR #1084, opened 2026-07-08
Bug 2053463 - Create dedicated worker pools for updatebot

ahal:
Updatebot will soon start using on task LLMs which makes it susceptible to prompt injection. Isolated pools ensures the blast radius is very small if it happens.

### DIFF
--- a/worker-pools.yml
+++ b/worker-pools.yml
@@ -436,6 +436,17 @@ pools:
         chain-of-trust: trusted
         suffix: ''
         instance_types: [*c3d-standard-16-lssd]
+      # Updatebot runs on-task LLMs so isolate to mitigate prompt injections
+      - pool-group: gecko-1
+        suffix: 'updatebot'
+        instance_types: [*c3d-standard-16-lssd]
+      - pool-group: gecko-2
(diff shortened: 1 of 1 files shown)


worker-pools.yml
```diff
@@ -2121,6 +2135,14 @@ pools:
       - pool-group: gecko-3
         chain-of-trust: trusted
         suffix: headless
+      # Updatebot runs on-task LLMs so isolate to mitigate prompt injection
+      - pool-group: gecko-1
```
bhearsum, line 2139: I saw it mentioned on Matrix that updatebot doesn't run on Try; are you sure we need L1 workers?

ahal: I'm not sure. AIUI they use an L2 project branch to test, so I'm pretty sure they need the l2 pools at least.

That said.. in theory every task can be run on Try if it's selected with `./mach try fuzzy` (it just might not succeed). I think it's worth keeping the L1 pools here just in case. In the event someone _does_ schedule it for some reason, a failure is better than it pending until the deadline

bhearsum (approved):

PR #1086, opened 2026-07-10
Bug 2053655 - hooks: trigger reviewer-selection based on routing key rather than payload.action

shtrom:


hooks.yml
```diff
@@ -301,11 +299,11 @@ project-engwf/reviewer-assignment-production:
   template_file: hook-templates/project-engwf/reviewer-assignment-production.yml
   bindings:
     - exchange: exchange/taskcluster-github/v1/pull-request
-      routing_key_pattern: '#.mozilla-firefox.#.opened'  # e.g., primary.mozilla-firefox.firefox.opened or primary.mozilla-firefox.firefox.synchronize
+      routing_key_pattern: '#.mozilla-firefox.#.opened'
```
shtrom: Note: this also matches `primary.mozilla-firefox.infra-testing.#`, which means PRs on infra-testing trigger the staging AND the prod hook.

I can live with that.

hooks.yml
```diff
@@ -301,11 +299,11 @@ project-engwf/reviewer-assignment-production:
...
     - exchange: exchange/taskcluster-github/v1/pull-request
       routing_key_pattern: '#.mozilla-firefox.#.reopened'
     - exchange: exchange/taskcluster-github/v1/pull-request
-      routing_key_pattern: '#.mozilla-firefox.#.synchronize'  # e.g., primary.mozilla-firefox.firefox.opened or primary.mozilla-firefox.firefox.synchronize
+      routing_key_pattern: '#.mozilla-firefox.#.synchronize'
```
ahal: Same, best to add the owner and repo here as well. E.g `primary.mozilla-firefox.firefox.synchronize` (no need for wildcards)

hooks.yml
```diff
@@ -197,7 +197,11 @@ project-engwf/reviewer-assignment-staging:
   template_file: hook-templates/project-engwf/reviewer-assignment-staging.yml
   bindings:
     - exchange: exchange/taskcluster-github/v1/pull-request
-      routing_key_pattern: '#.infra-testing.#'  # e.g., primary.mozilla-firefox.infra-testing.opened or primary.mozilla-firefox.infra-testing.synchronize
+      routing_key_pattern: '#.infra-testing.opened'
+    - exchange: exchange/taskcluster-github/v1/pull-request
+      routing_key_pattern: '#.infra-testing.reopened'
+    - exchange: exchange/taskcluster-github/v1/pull-request
+      routing_key_pattern: '#.infra-testing.synchronized'
```
ahal: There's a typo here (should be `synchronize`)

hooks.yml
```diff
@@ -197,7 +197,11 @@ project-engwf/reviewer-assignment-staging:
   template_file: hook-templates/project-engwf/reviewer-assignment-staging.yml
   bindings:
     - exchange: exchange/taskcluster-github/v1/pull-request
-      routing_key_pattern: '#.infra-testing.#'  # e.g., primary.mozilla-firefox.infra-testing.opened or primary.mozilla-firefox.infra-testing.synchronize
+      routing_key_pattern: '#.infra-testing.opened'
```
ahal: I'd add the Github owner to the routing key here too just in case some repo with the same name exists elsewhere:
```
primary.<org>.infra-testing.opened
```

ahal (changes requested):

hooks.yml
shtrom: The idea behind wildcards was to support projects, too. But it's not a thing at the moment, and it triggers for infra-testing, too, which is not ideal. Removed.

hooks.yml
shtrom: Which it does, in the thunderbird org, now (: But they don't want PR support just yet.

hooks.yml
ahal: Ah thanks for the explanation. We can add something back for projects in the future. Though they might exist on any org, so it might be better to explicitly add them here as needed. We'll see.

